PR #87, opened 2013-11-26
Successfully compiled for Windows

kevlened:
I combined some suggestions from others that had issues with Windows. I also made it a bit easier to identify a JDK_HOME and JRE_HOME.

Here's the system I compiled the attached binary on:
64-bit Windows 7 Ultimate
32-bit Python 2.7.6
32-bit Oracle JDK 1.7.0_25
MinGW

To run successfully in Windows:
Install a JDK and add a JAVA_HOME environment variable that points to the JDK folder
Install Cython from http://www.lfd.uci.edu/~gohlke/pythonlibs/#cython
Install jnius from the attached executable or build your own module from source

If you run into errors, try the following:
Add C:\Program Files (x86)\Java\jdk[YOUR JDK VERSION]\jre\bin\server to your path
Make sure you're using a 32-bit JDK
Log out of Windows and log back in

https://github.com/kevlened/pyjnius/releases/tag/v1.1

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -78,9 +78,14 @@ else:
         raise Exception('Unable to determine JRE_HOME')
     cpu = 'i386' if sys.maxint == 2147483647 else 'amd64'
 
+    if platform == 'win32':
+        incl_dir = join(jdk_home, 'include', 'win32')
+    else:
+        incl_dir = join(jdk_home, 'include', 'linux')
+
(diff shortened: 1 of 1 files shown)


vazudew:
The same issue exists for Windows 64... PyJnius is great but sadly, has no foolproof installation methods .

PR #101, opened 2014-02-08
Add pyobjc requirement to documentation

dustinfarris:
On OSX, I had to install PyObjC before I could successfully run `make` to install pyjnius.  This PR notates this requirement in the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,15 +26,16 @@ hello
 Usage on desktop
 ----------------
 
-You need a java JDK installed (openjdk will do), cython, and make to build it
+You need a [Java JDK][] installed (openjdk will do), [cython][], [pyobjc][] and make to 
+build it:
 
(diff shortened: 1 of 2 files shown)


akshayaurora:
There is absolutely no need to install pyobjc for installing pyjnius.

Steps are simple

pip install pyjnius --user
or
    pip install cython
    git clone http://github.com/kivy/pyjnius
    cd pyjnius
    make
    export PYTHONPATH=/path/to/pyjnius


Testing::

    python -c "import jnius"

If it complains about jre/jdk missing, click on `more info`, download `JavaForOSX` from apple site, not the one from oracle,  that won't work. Install `Java for OS X 2015 001`


That's it you should be ready to use jnius on your system.

"No need to install pyobjc"

PR #112, opened 2014-05-09
Fix tests

abrasive:
- set classpath to find NativeInvocationHandler
- fix signature checking test to pass when it throws the correct
  exception
- throw a distinguishable exception for bad signatures


dessant:
The pr conflicts with master, and tests were fixed in https://github.com/kivy/pyjnius/pull/163.

PR #114, opened 2014-05-16
Upstream tracking

barrywark:
Hi Everyone, I’d like to start pushing our pyjnius patches back upstream. This PR contains our updates to `setup.py` for building on OS X (with JDK 1.7+) and Windows. We’ve added Tox for multi-platform builds and Conda for easier binary deployment as well.

Looking forward to the discussion and to following with more patches in the future.


tito:
Hi,

We talked about this patch before, it cannot be accepted. I though we have closed it before, but no. Even if the intention is good and welcomed, such changes like the naming is a stop.

Patch should not change the name, author or things like that :/

barrywark:
@tito just wanted to get feedback. We've released this code under the new name to prevent conflict with the jnius package and added the authorship change to make it clear that this was a forked codebase, but we'd prefer change the name back before this patch is merged as well. Again, if the rest of the changes are acceptable, we'll resubmit with the name changes reverted. I propose we add our authorship contributions to the end of the author list.

barrywark:
I also realize that we forked before the change to the MIT license. We're happy to accept the new license for our contributions.

dessant:
You are welcome to resubmit without the conflicting changes, but we don't add every contributor's name to that string, as you've probably seen.

tshirtman:
well, rebasing to remove the irrelavant commits and pushing to the same branch should be good enough, feel free to reopen when you consider it clean.

Issue #96, opened 2014-01-15
Include JVM exception details in python JavaException


eugfeld:
Pyjnius throws JavaException when it detects a JVM exception. However python's JavaException does not contain any information about JVM exception that caused it. Even though pyjnius invokes JNI->ExceptionDescribe (which prints it to stderr), still at times it is hard to correlate and it would be a lot more informative to actually include at least JVM exception message in python exception. 

In my local shadow I have the following code (jnius_utils.py):

cdef void check_exception(JNIEnv *j_env) except *:

```
cdef jmethodID toString = NULL
cdef jstring e_string
cdef jboolean isCopy

cdef jthrowable exc = j_env[0].ExceptionOccurred(j_env)
if exc:
    j_env[0].ExceptionDescribe(j_env)
    j_env[0].ExceptionClear(j_env)

    toString = j_env[0].GetMethodID(j_env, j_env[0].FindClass(j_env, "java/lang/Object"), "toString", "()Ljava/lang/String;");
    e_string = j_env[0].CallObjectMethod(j_env, exc, toString);

    pystr = convert_jobject_to_python(
                    j_env, <bytes> 'Ljava/lang/String;', e_string)

    raise JavaException('JVM exception occured: ' + pystr)
```
##

clintharris:
+1 for merging this patch. Exceptions have meaning; not being able to properly handle them is a significant issue/shortfall with PyJNIus.

KeyWeeUsr:
Closed by #115.

PR #115, opened 2014-05-18
Issue 96 improved exception handling

Lenbok:
I initially applied just the patch that is described in #96, but then wanted to go further -- primarily so that on the python side I can distinguish between exception types and decide whether to locally handle the exception or whether to propagate the exception up (perhaps with exception rewrapping).


tito:
I didn't saw this PR, someone just pointed it in the mailing list. Great job!!

PR #126, opened 2014-08-18
Changes and adaptations for getting a working version in Python 2 and Python 3

pistolario:
General changes.
New dependency: future


dessant:
This is probably not needed after https://github.com/kivy/pyjnius/pull/164 has been merged. Please mention if it fixes other issues too.

PR #133, opened 2014-09-17
API for human-readable method signatures

chrisjrn:
This module is my attempt at making it easier to write proxies in PyJnius. 

It provides:
* `signature()` a method for constructing JNI method signatures from jnius.JavaClass objects and our own primitive type annotations
* `@with_signature`, a replacement for `@java_method`, which mimics the API of `signature()`
* tests to verify that the above work.

I'm opening this pull request as a request for comments, and to see if the approach is worthwhile (I certainly think it is). The unit tests (test_signatures.py) is probably the easiest way to see the approach in action.


alanjds:
+1

dessant:
:+1:

Issue #137, opened 2014-12-09
Class not found 'org/jnius/NativeInvocationHandler'


AdamAtri:
Tried to run "make tests":

adam@GPDevMachine:~/Downloads/tmp/pyjnius$ make tests
javac jnius/src/org/jnius/NativeInvocationHandler.java
python setup.py build_ext --inplace -f -g
/usr/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'install_requires'
  warnings.warn(msg)
running build_ext
cythoning jnius/jnius.pyx to jnius/jnius.c
warning: jnius/jnius.pyx:2:0: **getslice**, **setslice**, and **delslice** are not supported by Python 3, use **getitem**, **setitem**, and **delitem** instead
building 'jnius' extension
x86_64-linux-gnu-gcc -pthread -fno-strict-aliasing -DNDEBUG -g -fwrapv -O2 -Wall -Wstrict-prototypes -fPIC -g -I/usr/lib/jvm/java-7-openjdk-amd64/include -I/usr/lib/jvm/java-7-openjdk-amd64/include/linux -I/usr/include/python2.7 -c jnius/jnius.c -o build/temp.linux-x86_64-2.7/jnius/jnius.o
jnius/jnius.c: In function ‘__pyx_f_5jnius_get_jnienv’:
jnius/jnius.c:2933:3: warning: passing argument 2 of ‘(___pyx_v_5jnius_jvm)->AttachCurrentThread’ from incompatible pointer type [enabled by default](__pyx_v_5jnius_jvm[0])->AttachCurrentThread(__pyx_v_5jnius_jvm, (&__pyx_v_env), NULL);
   ^
jnius/jnius.c:2933:3: note: expected ‘void *_’ but argument is of type ‘const struct JNINativeInterface_ ***’
jnius/jnius.c: In function ‘__pyx_pf_5jnius_13MetaJavaClass_4resolve_class’:
jnius/jnius.c:13091:30: warning: assignment from incompatible pointer type [enabled by default](__pyx_v_jargs[0]) = ((jobject *)__pyx_v_classLoader);
                              ^
jnius/jnius.c:13100:30: warning: assignment from incompatible pointer type [enabled by default](__pyx_v_jargs[1]) = __pyx_v_interfaces;
                              ^
[...]

PR #138, opened 2015-01-03
Fix JVM signatures for the java.lang.Class methods.

tonyfinn:
The methods defined in the Class class contain semicolons at the end of the signatures for methods which return primitives, while these are only required after class names. As a result, a JVM error occurs when attempting to call any of the affected methods.


tonyfinn:
This failed the tests because the master branch was previously already failing the tests, as far as I can tell: https://github.com/kivy/pyjnius/issues/137

tito:
Looks good, thanks!

PR #149, opened 2015-02-26
OS X: use JavaVM framework from current Mac OS SDK

msmolens:
Fixes build for OS X 10.10 (Yosemite).
Build failed because JNI headers are missing in the framework path returned by `objc.pathForFramework('JavaVM.framework')`. This change  uses the framework from the current Mac OS SDK, as returned by `xcrun`, which includes the JNI headers.


tito:
Perfect. :)

PR #150, opened 2015-02-26
Enable setting primitive data type fields on Java classes

msmolens:
I encountered a Java library whose API requires setting public fields on a struct-like class instead of using setters/getters. This change adds support to Pyjnius for setting fields of primitive data types. Tests are added to BasicsTest.


tito:
Unsure about the usage, but looks good enough to be merged. Thanks :)

PR #144, opened 2015-01-29
Update jnius_conversion.pxi

retsyo:
change chr(<char>j_chars[i]) to unichr(j_chars[i]), now it can deal with Chinese

PR #153, opened 2015-03-13
Another missing PR

remram44:
#144 marked as merged but the changes were lost?

See #152. Be careful when force-pushing (or, preferably, just don't)


remram44:
This is still not in as of current master

remram44:
Bump. cc @retsyo (original author of #144)

akshayaurora:
This needs more analysis, why we are loosing these changes. I don't think any of us with push access force push.

PR #177, opened 2015-11-04
Followup fixes for python 3

benson-basis:
* stop using `future` except in tests, use `six` instead
* fix py2 vs py3 issues with `decode` in `setup.py`, might fix windows. Might fix android.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ before_install:
   - sudo apt-get install python-pip openjdk-7-jdk
 
 install:
-  - pip install --upgrade cython future six
+  - pip install --upgrade cython six
 
 script:
   - make
--- a/tests/test_assignable.py
+++ b/tests/test_assignable.py
@@ -1,8 +1,6 @@
 from __future__ import print_function
 from __future__ import division
 from __future__ import absolute_import
-from future import standard_library
-standard_library.install_aliases()
 import unittest
 from jnius import autoclass, JavaException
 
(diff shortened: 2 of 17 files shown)


benson-basis:
Android succeeds at create with these changes.

dessant:
We could avoid using six in setup.py, you could declare `PY3 = sys.version_info >= (3,0,0)` somewhere at the start and use that boolean for further checks.

dessant:
It does fail for windows even with pyhon 2, but i think that is not caused by your code, I've tried to install an older commit and it is just as broken. I'll post some error logs regarding that, but the priority is android, linux and osx for now.

.travis.yml
```diff
@@ -9,7 +9,7 @@ before_install:
   - sudo apt-get install python-pip openjdk-7-jdk
 
 install:
-  - pip install --upgrade cython future
+  - pip install --upgrade cython future six
```
akshayaurora: future not needed anymore?

akshayaurora:
@benson-basis thanks for this,  could we remove future dependency all together? instead of using both future and six?

benson-basis:
It's time-consuming to get all the 'future' out of the tests. I'll do it,
but I wanted to unblock things first.

benson-basis:
@dessant I think this should solve everything mentioned except for adroid and getting rid of future from the tests. As noted elsewhere, I'm a bit stumped on the android side.

inclement:
Thanks for the fixes. I've manually removed the apk stuff in the Makefile though as it's out of place (especially with the hardcoded machine-specific paths) - it would be better to add a test of some kind, or perhaps a pyjnius-specific test in p4a itself.

benson-basis:
Thanks.

PR #89, opened 2013-12-10
Add support for other include paths on other targets

thopiekar:
Will look at the jre_home to determinate which str() is needed for the variable, called "cpu".

PR #184, opened 2015-11-19
Adding support for armhf builds

thopiekar:
This is a rewritten PR of #89, which I wrote before.
While reading my old code I decided to code it another way.

I added a detailed description of what I've changed in the first commit. If you want more info, just ask, and feel free to give feedback :smile:


tito:
Looks good, thanks!

Issue #197, opened 2016-01-28
Import error on a clean install 


jsbruneau:
The current state of the tree is broken on windows 10

Last commit a00ce3d3908082f828663bf2b029b50320221a10

`Python 2.7.10 (default, May 23 2015, 09:40:32) [MSC v.1500 32 bit (Intel)] on win32
Type "help", "copyright", "credits" or "license" for more information.

> > > import jnius
> > > Traceback (most recent call last):
> > >   File "<stdin>", line 1, in <module>
> > >   File "jnius__init__.py", line 12, in <module>
> > >     from .jnius import *
> > > ImportError: No module named jnius`
##

jk1ng:
Hi jsbruneau, you might want to ensure that your PATH and JAVA_HOME environment variables are configured correctly.

For windows, JAVA_HOME should point to your JDK root; and PATH should include a path that goes to your jvm.dll file -- which should be in your JAVA_HOME subdirectory of /jre/bin/server

Hope that helps.

KeyWeeUsr:
Ref https://github.com/kivy/pyjnius/issues/199#issuecomment-443013378.

PR #200, opened 2016-02-07
setup fixes for python 3.5

danielepantaleone:
This PR fixes an issue that arises when building under python 3.5.

Issue #199, opened 2016-02-05
setup.py doesn't work with Python 3.5


jhidding:
Running "python setup.py install" with Python 3.5 gives the following error:
... AttributeError: 'str' object has no attribute 'decode' 
The bug is in the 'getenv' function, where the result of os.environ.get is being decoded. The error is only triggered in certain cases, installing on a clean system.
##

jk1ng:
I haven't verified this yet, but I suspect that os.environ returns unicode across multiple platforms. I've managed to install* it by editing out the .decode() call.

*Now I am having an issue with:

>     from .jnius import *
>     ImportError: DLL load failed: The specified module could not be found.

jk1ng:
I'd like to add that a workaround for my issue was ensuring that I had the right PATH and JAVA_HOME environment variable set. See Issue #197

KeyWeeUsr:
Two issues basically. One closed by #200 the `decode()`. Another one has a documentation section added via #361.

PR #202, opened 2016-02-09
Updated readme to counter issues #197

jk1ng:
Readme.md now tells users to ensure that their `PATH` and `JAVA_HOME` (or `JDK_HOME`) environment variables must be set properly in order to ensure the successful usage of PyJNIus. This should help counter issues #197 -- not the most automated fix, but still.

Apologies for the triple commit, apparently I don't know how to use markdown properly :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,11 @@ Usage on desktop
 ----------------
 
 You need a java JDK installed (OpenJDK will do), Cython and make to build it.
+Please ensure that your `JDK_HOME` or `JAVA_HOME` environment variable points
+to the installed JDK root directory, and that the JVM library (`jvm.so` or
+`jvm.dll`) is available from your `PATH` environment variable. **Failure to do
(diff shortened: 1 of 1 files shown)


jk1ng:
As an addendum, this doesn't solve #199 -- `danielepantaleone`'s PR #200 does that.

danielepantaleone:
You should [squash](http://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git) all those commits together to keep a clean git history.
Also I think the problem should be addressed at a certain point. I had the very same problem in the application I'm using PyJNIus in and I have added some code to automatically lookup the JVM path on different operating systems in case the JAVA_HOME and Path variables are not setup correctly. Something similar to what [JPype](https://github.com/originell/jpype/blob/master/jpype/_core.py#L94) is doing....not a very clean approach imho, but it would be better than nothing.

akshayaurora:
@jk1ng thanks

PR #215, opened 2016-03-30
Fix ImportError pyjnius: (missing -lpython3.5m flags)

haricot:
fix issue [658](https://github.com/kivy/python-for-android/issues/658)


kived:
This will break compilation on Python 3.4. This is an Android-specific problem, and should probably be fixed in python-for-android.

PR #221, opened 2016-05-19
Added import hook for java classes

inclement:
This doesn't actually work right, but if it did...would it be a good idea?


tito:
This is bad in my opinion, but wish to merge it only if you don't pre import the module in `jnius/__init__.py` :)

inclement:
I also think it's bad, so let's just close :)

Arnie97:
If anyone is still interested in this idea, I have created a similar module:
https://github.com/Arnie97/import-java

PR #271, opened 2017-05-08
for...in iteration of list interfaces

InvalidCo:
Throw index error instead of OutOfBoundsError, allowing us to use for...in with List interfaces.

https://docs.python.org/3.7/reference/datamodel.html#object.__getitem__


tito:
That's cool! thanks!

Issue #279, opened 2017-06-12
Windows 10 Home 64bit: Tests as well as minimum example fail with fatal JRE error 


hanslovsky:
I am not very familiar with windows but in order to distribute a python package that depends on PyJNIus on all major OSes (Windows, OSX, Linux) I tried to build PyJNIus on a Windows 10 Home Edition 64 bit machine.

I obtained python 3 through conda and followed (roughly) these steps:
```
 - Install conda
 - Install PyJNIus dependencies through conda and/or pip
 - Install OpenJDK `1.8.0_131`
 - install Apache ant
 - install visual cpp build tools
 - Instead of make run commands in Makefile by hand
   - `ant all`
   - `python setup.py build_ext --inplace -f`
   - for tests:
     - `cd tests`
     - `set CLASSPATH=..\build\test-classes;..\build\classes`
     - `set PYTHONPATH=..`
     - `nosetests -v > nosetests.log 2>&1` (compare nosetests.log in this gist)
```
The last command (nosetests) triggers this error message (I skipped the nosetests output but you can find that [here](https://gist.github.com/hanslovsky/69ab3d007c09749ce400b678829eea18#file-nosetests-log):
```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  EXCEPTION_ACCESS_VIOLATION (0xc0000005) at pc=0x00007fff4facdff4, pid=13156, tid=0x0000000000002f84
#
# JRE version: Java(TM) SE Runtime Environment (8.0_131-b11) (build 1.8.0_131-b11)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (25.131-b11 mixed mode windows-amd64 compressed oops)
# Problematic frame:
# C  [jnius.cp36-win_amd64.pyd+0x1dff4]
#
# Failed to write core dump. Minidumps are not enabled by default on client versions of Windows
#
# An error report file with more information is saved as:
# C:\Users\Zottel\Downloads\pyjnius-master\pyjnius-master\tests\hs_err_pid13156.log
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#
```

Similarly, when I try to implement a Java interface in Python and pass it to a Java method like this:
```python
# See fail.log for error message
import jnius_config
[...]

PR #290, opened 2017-08-11
Fixed issue #279 which causes callbacks to fail in Windows 64 bit

madhawav:
The issue #279 was caused due to invalid casting from c++ _long long_ to c++ _long_. This caused the value of variable _ptr_ of java class _NativeInvocationHandler_ become invalid (due to loss of precision by casting) causing _Access Violation Exception_ in Windows. The above exception is experienced during callbacks.

The issue was fixed by correcting cast _long_ to _long long_.


skalarproduktraum:
:+1:

akshayaurora:
@madhawav thanks. 👍

madhawav:
@akshayaurora 
when can we expect this fix to be available in PyPI?

hanslovsky:
Thanks for fixing and merging @madhawav @akshayaurora

PR #291, opened 2017-08-21
Fixed mishandling of long arguments in multiple methods. Added test cases.

jtoledo1974:
It was not possible to pass actual long values to multiple methods. It made it impossible to call AlarmManger.set in android, for instance.

Test cases are included.


tito:
Looks good, thanks you!

PR #285, opened 2017-07-04
Switching to JNI UTF-16 API for strings to fix supplementary unicode char encoding

brandonherzog:
Fix for #284

Switched all string creation/decoding to use the JNI UTF-16 api for unicode rather than try to wrangle with JNI's modified UTF-8 encoding.

I kept the existing assumption for Python 2 that byte strings are UTF-8 encoded.

PR #299, opened 2017-10-06
Fixes for Implementing a Java Interface (DONE BUT CANT REMOVE wip FLAG)

drmoose:
This PR fixes the following problems my team discovered while trying to implement a Java interface in Python:

* Forgetting to call `super().__init__` segfaults the JVM (#301) 
* On Python 3, `str` return values from `java_method`s don't get converted properly because the special case for `str` is in an elseif after the `basestring` case.
* Python's `isinstance` and `issubclass` do not work on Java classes & interfaces
* Method overload detection fails when trying to pass a `PythonJavaClass` to a function with the signature `(Ljava/lang/Object;)`

Written in collaboration with @kyonifer, @RatsRatsRats, @jacskiier.

### DIFF
--- a/tests/test_proxy.py
+++ b/tests/test_proxy.py
@@ -15,7 +15,6 @@ class TestImplemIterator(PythonJavaClass):
         'java/util/ListIterator', ]
 
     def __init__(self, collection, index=0):
-        super(TestImplemIterator, self).__init__()
         self.collection = collection
         self.index = index
 
--- a/jnius/__init__.py
+++ b/jnius/__init__.py
@@ -11,6 +11,7 @@ __version__ = '1.1.2-dev'
 
 from .jnius import *  # noqa
 from .reflect import *  # noqa
+from six import with_metaclass
 
 # XXX monkey patch methods that cannot be in cython.
 # Cython doesn't allow to set new attribute on methods it compiled
@@ -18,7 +19,7 @@ from .reflect import *  # noqa
 HASHCODE_MAX = 2 ** 31 - 1
 
 
-class PythonJavaClass_(PythonJavaClass):
+class PythonJavaClass_(with_metaclass(MetaJavaBase, PythonJavaClass)):
 
     @java_method('()I', name='hashCode')
     def hashCode(self):
--- a/jnius/jnius.pyx
+++ b/jnius/jnius.pyx
@@ -86,8 +86,8 @@ Python::
 '''
 
 __all__ = ('JavaObject', 'JavaClass', 'JavaMethod', 'JavaField',
-           'MetaJavaClass', 'JavaException', 'cast', 'find_javaclass',
-           'PythonJavaClass', 'java_method', 'detach')
+           'MetaJavaBase', 'MetaJavaClass', 'JavaException', 'cast',
+           'find_javaclass', 'PythonJavaClass', 'java_method', 'detach')
 
 from libc.stdlib cimport malloc, free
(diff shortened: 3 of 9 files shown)


drmoose:
After trying [very, very hard](https://github.com/drmoose/pyjnius/commit/79fcb29537d025d0c3626e97301d7962831e7de5) to find an environment I can run locally that'll reproduce the behavior seen above, I must concede defeat. The error output seems to have something to do with string encoding. 

eb4542b is the only place where we even touched encodings, and it was to fix errors very much like these (where strings get turned into "å".) I'm going to try force-pushing a version of the branch without that commit in it, just to see what your Travis does. Is it possible Travis's Python was compiled with weird/broken unicode support?

drmoose:
Same story. Weird encoding-related errors. As a further experiment, here's a force-push that's a single commit that adds one line of whitespace to the makefile and changes literally nothing else, in pursuit of the theory that Travis is just _broken_.

drmoose:
[Same errors, even with a commit that differs from master by a single blank line added to the Makefile](https://travis-ci.org/kivy/pyjnius/builds/284458383?utm_source=github_status&utm_medium=notification)

drmoose:
As the Travis build does not work no matter what I do, I have force-pushed one last time, restoring the original content of this PR. Perhaps on Monday I will look again and see whether I can figure out what's wrong with Travis, but I'm pretty sure at this point the problem is not anything to do with this PR.

inclement:
@drmoose I'm sorry for the delay in responding to this, the fixes look good but it seems nobody is very actively monitoring the pyjnius PRs. Don't worry about travis (I'll look into disabling or fixing it...), and I'll try to check through this soon.

drmoose:
It turns out #285 is a much better, much further-reaching fix for the encoding problem than just fixing the order the `if` branches like we did here. The [master of my fork](https://github.com/drmoose/pyjnius) resolves the merge conflict in favor of #285. I recommend merging #285 first & I'll rebase this PR.

tito:
Hi,

Could you update the PR to fix the conflict in jnius_conversion.pxi?

Thanks !

tests/test_proxy.py
```diff
@@ -15,7 +15,6 @@ class TestImplemIterator(PythonJavaClass):
         'java/util/ListIterator', ]
 
     def __init__(self, collection, index=0):
-        super(TestImplemIterator, self).__init__()
```
tito: Can you just explain me why this is not necessary? (i always call super)
Will it break current code available today that calls super() in init ?

drmoose: I removed that line to test what happens if the user forgets to call `super`. Having a super call there does not break anything, but removing it means f0cd0f4 isn't covered by tests.

drmoose:
Flagged WIP because post-rebase this no longer compiles in my test environment. It looks like I botched a merge conflict resolution somewhere.

drmoose:
Once again, I can't seem to reproduce the way Travis is failing. The error it's showing _looks_ like a legitimate problem with this PR, but I went through the same dockerized-testing script as before, and can't find an environment that fails that assertion.

drmoose:
I never did manage to reproduce Travis's error locally, but by repeatedly force-pushing on here I was able to get useful debug output from Travis and fix the problem (which was a missing special case). I believe this branch is now ready, and have un-WIP'd it for what is hopefully the last time. Sorry for the false starts.

PR #303, opened 2017-10-15
Fix bad JavaException in _getitem

psader:
Fix list iteration test broken since commit efe610b8, which results in a
JavaException with a bad message being thrown.  This was caused by the call to
'Class.forName', which throws the exception.


tito:
Nice one, thanks you

PR #310, opened 2017-11-20
Fix traceback.print_exc call

gdyuldin:
`traceback.print_exc` expects `limit` as a first argument. Passing
excepiton instance (AssertionError in example below) raises TypeError
inside of traceback module:
```
TypeError: unorderable types: AssertionError() >= int()
```


inclement:
Thanks.

PR #308, opened 2017-10-24
Add missing NativeInvocationHandler

drmoose:
This PR closes #288, #223, and #137, the various issues complaining that the NativeInvocationHandler is missing in the `pip install`ed version of pyjnius.

It cherry-picks @jessecrossen's [26f1367a](https://github.com/kivy/pyjnius/commit/26f1367ab143cd72203f0db77556279eac8c304f) referenced in #137 to include the NativeInvocationHandler's folder in the pip package and adjust the classpath accordingly, and adds a `subprocess.check_call` to `javac` to actually build the `.class` file.

Tested so far on:
- Ubuntu Xenial (openjdk8, py2)
- ArchLinux (openjdk8, py3)
- Windows 10 (oracle jdk 8, Anaconda Python 3.6.3)

PR #324, opened 2018-01-29
Cast path argument to list for classpath

djlambert:
Regression from #308. jnius_config.py line 72 is attempting to concatenate tuple.


djlambert:
Sorry, just noticed #319 also corrects this.

PR #325, opened 2018-01-30
Only remove 2 characters from field names when method begins with "is"

djlambert:
Field names generated from getter method names beginning with "is" have the first character chomped off (ie. ```Class.isEnabled()``` becomes ```Class.nabled```). This PR adds a check to determine slice size. Test case included.


kamathln:
Does this fix https://github.com/kivy/python-for-android/issues/945 ? I didn't know If I had to file it under pyjnius or python-for-android.

djlambert:
Any thought on getting this merged?

PR #329, opened 2018-02-18
Search for JRE in env vars

farizrahman4u:


tshirtman:
I can imagine this would break it for someone who had wrong JRE_HOME value before? Probably not a big concern, but i don't use pyjnius these days, and it could be an issue for some desktop users.

Issue #300, opened 2017-10-08
Strings Encoding Issues in Python 3 due to latest Cython updates


madhawav:
The latest release version of Cython (0.27.1) is causing pyjnius issues related to string encoding. I have experienced this issue in Python 3.5 for Ubuntu 16.04 (fresh installation). 

The issue is as follows. 
> A strings passed from Python to Java as parameter of a method gets encoded incorrectly. For example, when I pass the string "query" as a method parameter and _system.out.println_ from Java, it gets printed as "å".

When I install pyjnius using Cython 0.25.2, this issue is resolved. 
Also, this issue does not occur with Python 2.7.

iMcG33k:
The same issue on  Cython (0.27.1)  Python 3.6  macos 10.13 JDK 1.8.
But java String Works well
`
Stack = autoclass('java.util.Stack')
stack = Stack()
JString = autoclass('java.lang.String')
stack.push(JString("hello"))
stack.push(JString("world"))
print(stack.pop())
`
output:
`
Hello world

world
hello
`

psader:
The issue is resolved by modifying `populate_args` in `jnius_conversion.pxi`, such that the input `py_str` is always converted with `.encode('utf-8')` before calling `NewStringUTF`. This means only running the code under `except (UnicodeEncodeError, TypeError):` so that there's only one possible call to `NewStringUTF`.
The test drivers pass with this change with python 2 and 3. There may be some case not covered by them, but calling a string conversion routine twice to see which one works is a bad code smell, because otherwise you end up with the "Bush hid the facts" bug (https://en.wikipedia.org/wiki/Bush_hid_the_facts).

psader:
There's also https://github.com/kivy/pyjnius/pull/285 to resolve this.

jakirkham:
Should this be closed now that PR ( https://github.com/kivy/pyjnius/pull/285 ) is merged or is this still outstanding?

people-can-fly:
one workaround from @Ariel-Lieberman using `java.long.string`

```
Stack = autoclass('java.util.Stack')
jString = autoclass('java.lang.String')
stack = Stack()
stack.push(jString('OK'))
print(stack.pop())
```

KeyWeeUsr:
Resolved by #285. If the issue still persists, open a **new** issue and reference old ones if they are related. This one is rather messy together with the references.

PR #334, opened 2018-03-06
fix: readme reflecting right fragment about java.lang.String

akkumar:
The example needs explicit instantiation of java.lang.String , at least with Java8. 

So - needs to be modified per se.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,8 +27,9 @@ Hello world
 
 >>> Stack = autoclass('java.util.Stack')
 >>> stack = Stack()
->>> stack.push('hello')
->>> stack.push('world')
+>>> String = autoclass('java.lang.String')
+>>> stack.push(String('hello'))
+>>> stack.push(String('world'))
 >>> print stack.pop()
 world
 >>> print stack.pop()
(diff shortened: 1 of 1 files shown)


Arnie97:
At least not so on my PC. Anyone else got this bug?

```
Python 3.6.2 (v3.6.2:5fd33b5, Jul  8 2017, 04:57:36) [MSC v.1900 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
```
```python
>>> from jnius import autoclass
>>> System = autoclass('java.lang.System')
>>> System.getProperties().get('java.version')
'1.8.0_152-release'
>>> Stack = autoclass('java.util.Stack')
>>> stack = Stack()
>>> stack.push('hello')
'hello'
>>> stack.push('world')
'world'
>>> print(stack.pop())
world
>>> print(stack.pop())
hello
```

Arnie97:
Are you trying to provide a workaround for #300?

akkumar:
@Arnie97 - That sounds about right regarding issue with cython. 

Yes - I was using Cython==0.27.3. Reproducible with Cython==0.28.2 as well. (with python 3.6.5 )

```bash
$ java -version
java version "1.8.0_162"
Java(TM) SE Runtime Environment (build 1.8.0_162-b12)
Java HotSpot(TM) 64-Bit Server VM (build 25.162-b12, mixed mode)
```

For what it is worth - here is 
```python
>>> from jnius import autoclass
>>> System = autoclass('java.lang.System')
>>> System.getProperties().get('java.version')
None
```

akkumar:
Given that a workaround exists as suggested by @Arnie97 - closing it hence. 

Now - it would be great to have a release with the patch though.

PR #336, opened 2018-04-23
resolve_class: use bytes default objects when required

abrasive:
Fixes a crash on Python 3.5.5 where `PyBytes_Check()` does not accept the
str() objects previously supplied as defaults for `__javainterfaces__` and so on. I'm not sure if this is a recent Python change or what.

Crash looks like:
```
jnius/jnius.c:17873: __pyx_pf_5jnius_5jnius_13MetaJavaClass_4resolve_class: assertion "PyBytes_Check(__pyx_v___javainterfaces__)" failed'
```

This is the cause of kivy/python-for-android#1247.

### DIFF
--- a/jnius/jnius_export_class.pxi
+++ b/jnius/jnius_export_class.pxi
@@ -57,8 +57,8 @@ class MetaJavaClass(type):
 
         cdef JavaClassStorage jcs = JavaClassStorage()
         cdef bytes __javaclass__ = <bytes>classDict['__javaclass__']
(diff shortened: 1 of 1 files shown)


AndreMiras:
It looks good @abrasive, I'll give it a try next week when I've access to my Gentoo.
In fact I think Ubuntu also has that issue, but assert are disabled with `-DNDEBUG`, but on Gentoo this preprocessor flag got ignored for some reason.

AndreMiras:
Hi @abrasive I could give it a try on Gentoo and I confirm it works.
It was with Python3.6 host and Python3.6/Crystax target.

PR #347, opened 2018-07-18
Fix class name generated by autoclass

JPDSousa:
This is just a minor fix. It is important to show the determinism of generated class names.


JPDSousa:
To give additional context: classes generated through `jnius.autoclass` follow a naming strategy of `jnius.reflect` + `fully qualified java class name`. This rule is used for all classes, as far as I've tested. At least, in case of the example displayed `java.lang.System`, the generated python class is actually `jnius.reflect.java.lang.System`.

tshirtman:
Thanks!

PR #352, opened 2018-09-04
Changed JRE detection on Windows.

Taikamya:
This added Windows 7/8/10 compatibility with JRE/JDK 10 and 11.

Fixed some typos. ie: "amd64" to "AMD64"
Fixed some wrong calls. ie: "machine2cpu.keys()" to "machine2cpu.values()".
Fixed some redundancy.
Fixed some indentation.
Fixed old compatibility and added new ones.

Reformatted a small part of the code to make it more Pythonic.

Tested and working well on Windows 7, Python 3.6 both versions.
Tested and working well on Windows 10, Python 3.7 both versions.

Signed-off-by: Dimitri Moreira <taikamya@outlook.com>

PR #366, opened 2018-11-24
On OSX we should favor Java distribution that declare JAVA_HOME

hadim:
A last PR to favor JAVA_HOME during OSX build.

I need this to fix the feedstock build since OSX Travis image comes with a preinstalled Java package and I need to use the Java versions shipped by Conda.

Ping @KeyWeeUsr

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -102,22 +102,22 @@ if PLATFORM == 'android':
 
 elif PLATFORM == 'darwin':
 
-    JAVA_HOME = subprocess.Popen(
+    FRAMEWORK = subprocess.Popen(
         '/usr/libexec/java_home',
         stdout=subprocess.PIPE, shell=True).communicate()[0]
 
     if PY3:
-        JAVA_HOME = JAVA_HOME.decode()
+        FRAMEWORK = FRAMEWORK.decode('utf-8')
 
-    JAVA_HOME = JAVA_HOME.strip()
+    FRAMEWORK = FRAMEWORK.strip()
 
-    if not JAVA_HOME:
+    if not FRAMEWORK:
         raise Exception('You must install Java on your Mac OS X distro')
 
-    if '1.6' in JAVA_HOME:
+    if '1.6' in FRAMEWORK:
         LIB_LOCATION = '../Libraries/libjvm.dylib'
         INCLUDE_DIRS = [join(
-            JAVA_HOME, (
+            FRAMEWORK, (
                 'System/Library/Frameworks/'
                 'JavaVM.framework/Versions/Current/Headers'
             )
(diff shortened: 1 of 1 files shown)


hadim:
@KeyWeeUsr do you have an ETA for a new release? If you plan a lot of fixes before, would you mind release one before so I can start updating the conda forge feedstock? That would be really helpful!

hadim:
Only a git tag would be sufficient actually (no need for a pypi upload).

setup.py
```diff
@@ -99,41 +99,51 @@ def compile_native_invocation_handler(*possible_homes):
     # for android, we use SDL...
     LIBRARIES = ['sdl', 'log']
     LIBRARY_DIRS = ['libs/' + getenv('ARCH')]
+
 elif PLATFORM == 'darwin':
-    FRAMEWORK = subprocess.Popen(
```
KeyWeeUsr: I think the renaming will just cause confusion. Leave it as `FRAMEWORK`, please.

setup.py
```diff
@@ -99,41 +99,51 @@ def compile_native_invocation_handler(*possible_homes):
...
+    JAVA_HOME = subprocess.Popen(
         '/usr/libexec/java_home',
         stdout=subprocess.PIPE, shell=True).communicate()[0]
+
     if PY3:
-        FRAMEWORK = FRAMEWORK.decode()
-    FRAMEWORK = FRAMEWORK.strip()
-    print('java_home: {0}\n'.format(FRAMEWORK))
-    if not FRAMEWORK:
+        JAVA_HOME = JAVA_HOME.decode()
```
KeyWeeUsr: Decoding explicitly with `utf-8` seems as a better idea to me instead of relying on the defaults, even if this isn't section running on Windows.

KeyWeeUsr:
> 
> 
> @KeyWeeUsr do you have an ETA for a new release? If you plan a lot of fixes before, would you mind release one before so I can start updating the conda forge feedstock? That would be really helpful!

Sorry, no ETAs. I'll try to look at some `pyjnius` and `plyer` related issues and I'll see then. :)

hadim:
Would that make things faster if I setup Windows CI for the repo? Tell me whether you want to use Travis or Appveyor and I'll see what I can do.

KeyWeeUsr:
@hadim definitely yes! :+1: We'd need to check whether it can find both 32bit and 64bit JDK. Appveyor is used for Kivy + sister projects, so let's do that, since there already is an integration with it for GitHub PRs.

There is https://github.com/kivy/pyjnius/pull/352 with an approach for making the detection work for JDK properly (the current state relies on quotes not being present in the env variable, manual string slicing and similar bad stuff). The author doesn't respond though, so either feel free to pick the PR and test with it to preserve the commits or you can came with own approach. Anyway, if the state of the PR won't move until this weekend, I'll close it and come with a new one.

In case you'd want to pick the PR together with Appveyor integration, we'll need to check mainly the [preinstalled JDKs](https://www.appveyor.com/docs/windows-images-software/#java). Feel free to check with conda JDKs if you favor that one, but we'll mainly need the JDKs from the official location (i.e. `Program Files`).

KeyWeeUsr:
Thank you! :dango:

PR #367, opened 2018-11-27
Don’t install setup script to site-packages

flying-sheep:


KeyWeeUsr:
The setup.py is needed for source dist in this particular file. Sdist generates tar.gz which is then downloaded via pip and the setup.py inside of that tar.gz is run to actually compile and link against java.

flying-sheep:
Then solve that differently. The way you do things now installs `setup.py` to `site-packages`.

That means after installing pyjnius I can to `import setup` which runs your setup.py anywhere on my system. That’s unacceptable.

KeyWeeUsr:
>Then solve that differently. The way you do things now installs setup.py to site-packages.
>
>That means after installing pyjnius I can to import setup anywhere on my system which tries to run your setup.py. That’s unacceptable.

What about trying to explain the problem you're attempting to fix with the pull request **first** instead of an empty description or basically at least try to run the proposed fix first instead of `That's unacceptable` attitude and blindly removing stuff from setups?

flying-sheep:
Oh, of course, I’d have updated the PR if you hadn’t closed it!

PR #369, opened 2018-11-27
Remove setup.py from installed package

### LINKED ISSUE #368: Stop installing setup.py globally
When I install pyjnius via `python setup.py install`, the following line results in setup.py ending up directly in `site-packages`

https://github.com/kivy/pyjnius/blob/8aa91781f6d7726f9eb04fb8b1fb913be0ba1f4c/setup_sdist.py#L26

This is unacceptable misbehavior for a python package. I tried to remove it in #367, which got closed because it breaks your build. Fair enough, then please make sure another way that `setup.py` *only* ends up in the sdist and never in `site-packages`.

KeyWeeUsr:
Closes #368, ref #367.


flying-sheep:
Thank you!

PR #377, opened 2018-12-15
Doc tweaks

Zen-CODE:


docs/source/installation.rst
```diff
@@ -50,30 +50,30 @@ pyjnius to the requirements argument when creating a dist or apk::
...
+1. Download and install the JDK containing the JRE:
 
    http://www.oracle.com/technetwork/java/javase/downloads/index.html
 
 2. Edit your system and environment variables (use the appropriate Java bitness
-   and version in paths):
+   and version in the paths):
 
-    Add to `Environment Variables
+    Add to your `Environment Variables
```
AndreMiras: I know you didn't add that one, but what is the backtick for here? If it's a typo maybe you could remove it in this PR, what do you think?

AndreMiras: oh yes true for the link, thanks for the head up. I'm often getting confused with the rst syntax

Issue #385, opened 2019-01-01
Sending None/null in an argument of type Object


AbdealiJK:
I'm trying to call a function which has the signature:
```java
	 * @param value The input value in user-supplied representation.
	 * Use <code>null</code> to represent a missing input value.
	 */
	public FieldValue prepare(Object value){
		return ...
	}
```

To call it, when I do:
```
.prepare(600)
```
It throws an error: `JavaException: Invalid python object for this argument. Want 'java/lang/Object', got 600`

Is there any helper function I can use to convert my python variable to a Object?

For a temp workaround I needed to use `float`, `str`, `None`. So ... I tried:
```python
if isinstance(val, float):
    val = jnius.autoclass('java.lang.Double')(val)
elif isinstance(val, str):
    val = jnius.autoclass('java.lang.String')(val)
else:
    raise NotImplementedError("Unknown type: {}".format(type(val))
```

But I couldn't figure how to send `None` to this function which seems to support only `Object` as the argument.
The function does support `null` as a valid input. Unsure how to send it from jnius though.

KeyWeeUsr:
Fixed by #386.

PR #387, opened 2019-01-03
Add test for Python to Java function arguments conversion

KeyWeeUsr:
For issues such as #385, #386 and related.

### DIFF
--- a/jnius/jnius.pyx
+++ b/jnius/jnius.pyx
@@ -95,6 +95,7 @@ from functools import partial
 import sys
 import traceback
 
+include "jnius_compat.pxi"
 include "jni.pxi"
 include "config.pxi"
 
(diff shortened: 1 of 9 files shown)


KeyWeeUsr:
So, the test case with `JavaException` was an incorrect one, I fixed it. Also Cython doesn't handle `basestring` properly and has a differences in Python 2 and Python 3 runtime. Meaning:

    cdef void stuff(basestring arg): pass

will not accept this call: `stuff(b'')` with Python 3 and will happily pass with Python 2. Then there is also (most likely the same bug) an issue with:

    isinstance(arg, basestring)

which is correct with Python 2 when using both `bytes/str` and `unicode`, but is broken for Python 3 and basically everywhere we actually try to use `basestring` as an `elif` case it'll drop us right to the `else`. I hope we don't have much of such hidden surprises remaining hmmm.

PR #393, opened 2019-01-31
Include signatures to exceptions

KeyWeeUsr:
Currently the behavior in case of a mismatch of a signature is rather bad because from the exception we don't know what actually happened, therefore we guess. That's not good at all.

### DIFF
--- a/jnius/jnius_export_class.pxi
+++ b/jnius/jnius_export_class.pxi
@@ -74,7 +74,8 @@ class MetaJavaBase(type):
 
             if NULL != proxy and 0 != j_env[0].IsInstanceOf(j_env, obj, proxy):
                 # value is a proxy object. check whether it's one of ours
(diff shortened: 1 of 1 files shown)


jnius/jnius_export_class.pxi
```diff
@@ -410,16 +421,20 @@ cdef class JavaField(object):
...
             self.j_field = j_env[0].GetFieldID(
-                    j_env, self.j_cls, <char *>namestr,
-                    <char *>defstr)
+                j_env, self.j_cls, <char *>namestr,
+                <char *>defstr
+            )
         if self.j_field == NULL:
-            raise JavaException('Unable to found the field {0}'.format(self.name))
+            raise JavaException(
+                'Unable to found the field {0}'.format(self.name)
```
tshirtman: missed opportunity to fix "found" into "find" :smile:

PR #399, opened 2019-02-01
Get JDK/JRE path pythonically

### LINKED ISSUE #100: Error installing on OSX
`pip install jnius` fails on my OSX.

View the traceback on pastebin: http://pastebin.com/fvsDdRda

It appears that setup.py is [running a command](https://github.com/kivy/pyjnius/blob/master/setup.py#L66-L67), `readlink`, in a way that OSX doesn't like.  Here is my own output from running the same command:

``` console
$ readlink -f `which javac`
readlink: illegal option -- f
usage: readlink [-n] [file ...]
```

What is the intention of the -f flag on other operating systems?
## 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/2653467-error-installing-on-osx?utm_campaign=plugin&utm_content=tracker%2F77133&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F77133&utm_medium=issues&utm_source=github).
</bountysource-plugin>

KeyWeeUsr:
Remove `readlink -f` and `sed`, we have Python for replacing not `sed` + pipe and `os.readlink()` should work properly as a `readlink -f` alternative from Python's side.

Closes #100.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: generic
 
 os:
-  - osx
   - linux
+  - osx
 
 env:
   - PYTHON_VERSION="2.7" JAVA_VERSION="8"
--- a/setup.py
+++ b/setup.py
@@ -169,11 +169,8 @@ else:
             JDK_HOME = realpath(
                 subprocess.check_output(
                     ['which', 'javac']
-                ).strip()
+                ).decode('utf-8').strip()
             ).replace('bin/javac', '')
-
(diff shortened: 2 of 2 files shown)


KeyWeeUsr:
Ok, this is silly, because I can't get `which java(c)` to work properly on Linux. I will resolve this later after more important stuff is done.

PR #400, opened 2019-02-02
Convert java.lang.Long to INT if < 2 ** 31 else to LONG

KeyWeeUsr:
Convert java.lang.Long to INT if < 2 ** 31 else to LONG

LONG is in Py3 merged with INT type and renamed to INT without 'L' being present in the string representation of the number but having the full storage capacity of LONG. On Py2 however it will cause type issues when we convert java.lang.Long with size < 2 ** 31 into Py2 long and convert it back. Therefore convert java.lang.Long to Py2 long *only* if it exceeds the maximum value of INT (2 ** 31).

### DIFF
--- a/jnius/jnius_conversion.pxi
+++ b/jnius/jnius_conversion.pxi
@@ -170,7 +170,11 @@ cdef convert_jobject_to_python(JNIEnv *j_env, definition, jobject j_object):
     if r == 'java/lang/Long':
         retclass = j_env[0].GetObjectClass(j_env, j_object)
         retmeth = j_env[0].GetMethodID(j_env, retclass, 'longValue', '()J')
-        return j_env[0].CallLongMethod(j_env, j_object, retmeth)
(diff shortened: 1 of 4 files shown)


jnius/jnius_conversion.pxi
```diff
@@ -170,7 +170,11 @@ cdef convert_jobject_to_python(JNIEnv *j_env, definition, jobject j_object):
     if r == 'java/lang/Long':
         retclass = j_env[0].GetObjectClass(j_env, j_object)
         retmeth = j_env[0].GetMethodID(j_env, retclass, 'longValue', '()J')
-        return j_env[0].CallLongMethod(j_env, j_object, retmeth)
+        value = j_env[0].CallLongMethod(j_env, j_object, retmeth)
```
tito: You're calling at least twice the java method. So this not ok :)

Issue #188, opened 2015-12-08
Build non-trivial because of Cython


blootsvoets:
Currently, users need to install Cython before installing the package (in a separate command). This can be slightly frustrating for automatic installs. If the c-files are added to the repository, users could compile those with distutils, without needing Cython at all. This setup is advised in
http://docs.cython.org/src/reference/compilation.html#distributing-cython-modules

Would that be feasible for this package?
##

inclement:
I don't see a problem with doing this. I'm not familiar with best practice here, should the .c files perhaps only be in release versions, to avoid cluttering the repo and needing to keep them up to date?

blootsvoets:
Possible approach for only doing this during releases:
http://stackoverflow.com/a/18418524/574082

dessant:
Wheels are needed. :)

blootsvoets:
Would fix #208.

kived:
As you would still need a compiler, I fail to see how this would really make build less non-trivial. It wouldn't necessarily fix #208, and that's not a real issue anyway but something specific to that user's environment - that user has messed up his Python installation, most likely by assuming that you can use py3.4's site-packages dir with py3.5. Just because the immediate issue is the cython import doesn't mean there won't be a ton of other issues if that one is bypassed.

Tomaat:
Is there any progress on this issue? Because I now have to install Cython before pyjnius, and have a project with pyjnius as a dependency, I also have to install Cython before installing my project (which makes deploying harder)
Is it not possible to add Cython to the INSTALL_REQUIRES in setup.py?
e.g. setup.py:64-78
```python
try:
    from Cython.Distutils import build_ext
    INSTALL_REQUIRES.append('cython')
except ImportError:
    # pylint: disable=ungrouped-imports
    try:
        from setuptools.command.build_ext import build_ext
    except ImportError:
        from distutils.command.build_ext import build_ext
    if PLATFORM != 'android':
        print('\n\nYou need Cython to compile Pyjnius.\n\n')
        raise
    # On Android we expect to see 'c' files lying about.
    # and we go ahead with the 'desktop' file? Odd.
    FILES = [fn[:-3] + 'c' for fn in FILES if fn.endswith('pyx')]
```
would become something like
```python
try:
    from setuptools.command.build_ext import build_ext
except ImportError:
    from distutils.command.build_ext import build_ext

if PLATFORM != 'android':
    INSTALL_REQUIRES.append('cython')
else:
[...]

PR #417, opened 2019-04-29
setup.py: Add cython to setup_requires

AbdealiJK:
setup_requires is a method to add libraries that are needed during
the setup.py execution. This is picked up by easy_install and is
installed in the build directory for the build to occur.

This simplifies the installation of pyjnius from source
distributions as the cython dependency for build is managed by the
python package management toolchain.

This PR resolves https://github.com/kivy/pyjnius/issues/188 to some extent. While wheels are the ideal solution for it - this makes it simpler without much effort.
Other projects using setup_requires:
 - https://github.com/apache/arrow/blob/master/python/setup.py#L584
 - https://github.com/pandas-dev/pandas/blob/master/setup.py#L40

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -53,6 +53,7 @@ LIB_LOCATION = None
 EXTRA_LINK_ARGS = []
 INCLUDE_DIRS = []
 INSTALL_REQUIRES = ['six>=1.7.0']
+SETUP_REQUIRES = []
 
 # detect Python for android
 PLATFORM = sys.platform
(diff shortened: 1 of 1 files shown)


AbdealiJK:
@KeyWeeUsr - it would be awesome if you can take a look at this

AbdealiJK:
Thanks @tshirtman 
Looking forward to this in the nest release to simplify my devops a lot!

PR #420, opened 2019-05-15
Avoid windows execution error when JAVA_HOME path has space (test on …

jrnp97:
On windows installation when JAVA_HOME path has a space on string python os module make path with '\\' symbol and space in path made trigger window execution error 

Example below

image

Then to support command execution on windows system, we need to put path between "" 

image

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -84,6 +84,8 @@ def find_javac(possible_homes):
     for home in possible_homes:
         for javac in [join(home, name), join(home, 'bin', name)]:
             if exists(javac):
+                if sys.platform == "win32": # Avoid path space execution error
+                    return'"%s"' % javac
                 return javac
     return name  # Fall back to "hope it's on the path"
 


setup.py
```diff
@@ -84,17 +84,25 @@ def find_javac(possible_homes):
...
-        join('jnius', 'src', 'org', 'jnius', 'NativeInvocationHandler.java')
-    ])
+    try:
+        subprocess.check_call([
+            javac, '-target', '1.6', '-source', '1.6',
+            join('jnius', 'src', 'org', 'jnius', 'NativeInvocationHandler.java')
+        ])
+    except FileNotFoundError:
+        subprocess.check_call([
+            javac.replace('""', ''), '-target', '1.6', '-source', '1.6',
```
tshirtman: wouldn't it be better to do this replacement all the time? there is no sense in keeping an empty quote in the initial path, no? and it would look better than try/except.

setup.py
```diff
@@ -84,17 +84,25 @@ def find_javac(possible_homes):
     for home in possible_homes:
         for javac in [join(home, name), join(home, 'bin', name)]:
             if exists(javac):
+                if sys.platform == "win32" and not PY2: # Avoid path space execution error
+                    return'"%s"' % javac
```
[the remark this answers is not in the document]
jrnp97: Thks

setup.py
jrnp97: Sorry this replace suppose to be **.replace('"', '')** instead the version commented, the try-except is for try perform subprocess method on the old way on some system does not require add **"** (e.g when use powershell)

jrnp97: I understand your point, let's try to do on a better way

Issue #419, opened 2019-05-09
Inspect Java method arguments and return type


JPDSousa:
I'm working on a python tool on top of pyjnius. One of my technical requirements is to preprocess python objects based on the meta-information of a given java method.

For this I tried to access this data through `JavaMethod` (e.g. `object.hashCode` will return a JavaMethod). The thing is that I can't seem to access any information of the java method through `JavaMethod`. By doing `dir(object.hashCode)` I only get python related stuff.

Now, by inspecting the java method class, and bear with me as I'm not a cython expert, I see that I have a field [here](https://github.com/kivy/pyjnius/blob/master/jnius/jnius_export_class.pxi#L678) which points me to the fact that if what I want is not available today, it can be achieved.

Can I get some more insights on this? I might be able to help with a merge request.

amsukdu:
Hello sir,
I think this is a great idea and want to solve this issue. 
Poked around the source code and it seems that I can add the dunders(__dir__, __repr__, __str__, etc...) to JavaMethod python object.

My opinion is that __repr__ or __str__ will be more appropriate to express signature than __dir__. 
Secondly, the signature will be more informative if it is Java type like "method(String, Integer) -> String".

What are your thoughts? Please let me know! 😄

JPDSousa:
I agree that we might want to change `str` and `repr`, but since both of those methods will return a string, iterating over the list of arguments will be a bit trickier. My idea is to have that `definition_args` field accessible from the Python instance in the form of an array of `JavaClass` instances, or something similar.

What do you think? Does this match your expectations?

amsukdu:
Yeah, you're right. One huge string would be troublesome.

```python
from jnius import autoclass
cls = autoclass('java.lang.String')
dir(cls.substring)
#['(I)Ljava/lang/String;', '(II)Ljava/lang/String;']
dir(cls.valueOf)
'''
['(C)Ljava/lang/String;',
 '(D)Ljava/lang/String;',
 '(F)Ljava/lang/String;',
 '(I)Ljava/lang/String;',
 '(J)Ljava/lang/String;',
 '(Ljava/lang/Object;)Ljava/lang/String;',
 '(Z)Ljava/lang/String;',
 '([C)Ljava/lang/String;',
 '([CII)Ljava/lang/String;']
'''
```

The expression is from JNI(I think...).

JPDSousa:
[...]

PR #423, opened 2019-05-24
add dunder __dir__ for java methods

amsukdu:
For readable / usable propose, added `__dir__` to JavaMethod & JavaMultipleMethod. 
It is more human readable than JNI signatures.

Solves #419 issue
`readable_sig` is in .pxi but if it can be moved to .py anytime. I can make some test cases also.

### DIFF
--- a/tests/test_dir.py
+++ b/tests/test_dir.py
@@ -1,0 +1,43 @@
+from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
+import unittest
+from jnius.reflect import autoclass
+
+class DirTest(unittest.TestCase):
+
+    def test_varargs_dir(self):
+        cls = autoclass("java.lang.System")
+        assert isinstance(dir(cls.out.printf), list)
+        #[(['java/lang/String', 'java/lang/Object...'], 'java/io/PrintStream'),
+        # (['java/util/Locale', 'java/lang/String', 'java/lang/Object...'], 'java/io/PrintStream')]
+        for f in dir(cls.out.printf):
+            assert isinstance(f, tuple)
+
+    def test_array_dir(self):
+        cls = autoclass("java.util.List")
+        assert isinstance(dir(cls.toArray), list)
+        #[([], 'java/lang/Object[]'),
+        # (['java/lang/Object[]'], 'java/lang/Object[]')]
+        for f in dir(cls.toArray):
+            assert isinstance(f, tuple)
+
+    def test_dir(self):
+        cls = autoclass("java.lang.String")
+        assert isinstance(dir(cls.valueOf), list)
+        # [(['boolean'], 'java/lang/String'),
+        #  (['char'], 'java/lang/String'),
+        #  (['char[]'], 'java/lang/String'),
+        #  (['char[]', 'int', 'int'], 'java/lang/String'),
+        #  (['double'], 'java/lang/String'),
+        #  (['float'], 'java/lang/String'),
+        #  (['int'], 'java/lang/String'),
+        #  (['java/lang/Object'], 'java/lang/String'),
+        #  (['long'], 'java/lang/String')]
+        for f in dir(cls.charAt):
+            assert isinstance(f, tuple)
+
+
+
+
+
--- a/tests/test_dir.py
+++ b/tests/test_dir.py
@@ -7,24 +7,32 @@ from jnius.reflect import autoclass
 class DirTest(unittest.TestCase):
 
     def test_varargs_dir(self):
+        # >>> from jnius import autoclass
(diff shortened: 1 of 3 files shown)


JPDSousa:
@amsukdu can we see the output of the `dir` methods? Maybe include it as a comment on the review. I'll try to test it out on my end anyway. Also, I'm not aware of the standards for this repo, but maybe some automatic testing is in order as well :)

amsukdu:
@JPDSousa 
Yes, I can make some tests. I'll push it as soon as I wrote.

tshirtman:
That looks cool! Sorry for very late feedback, i do think this is valuable, if you can, a few tests for the helper function that converts from jni to java notation, would be nice, i'll likely merge either way, and write them myself if you don't have time for it.

amsukdu:
Hello, @tshirtman
I added more comments & fixed typos. I'm not 100% sure about what the helper function is referring to but I write the function `readable_sig` in cpython but it is basically the same as normal python code except `cdef`.

Your changes are welcome!

tests/test_dir.py
```diff
@@ -0,0 +1,53 @@
...
+from jnius.reflect import autoclass
+
+class DirTest(unittest.TestCase):
+
+    def test_varargs_dir(self):
+        # >>> from jnius import autoclass
+        # >>> cls = autoclass('java.lang.System')
+        # >>> dir(cls.out.printf)
+        # [(['java/lang/String', 'java/lang/Object...'], 'java/io/PrintStream'),
+        # (['java/util/Locale', 'java/lang/String', 'java/lang/Object...'], 'java/io/PrintStream')]
```
tshirtman: why not test that this is true, in the test? maybe
```
assert (['java/lang/String', 'java/lang/Object...'], 'java/io/PrintStream') in dir(cls.out.printf)
```

etc?

tests/test_dir.py
```diff
@@ -0,0 +1,53 @@
...
+        #  (['double'], 'java/lang/String'),
+        #  (['float'], 'java/lang/String'),
+        #  (['int'], 'java/lang/String'),
+        #  (['java/lang/Object'], 'java/lang/String'),
+        #  (['long'], 'java/lang/String')]
+        cls = autoclass("java.lang.String")
+        assert isinstance(dir(cls.valueOf), list)
+        for f in dir(cls.charAt):
+            assert isinstance(f, tuple)
+
```
tshirtman:
```suggestion
```

tests/test_dir.py
```diff
@@ -0,0 +1,53 @@
...
+
```
tshirtman:
```suggestion
```

tests/test_dir.py
```diff
@@ -0,0 +1,53 @@
...
+
```
tshirtman:
```suggestion
```

tests/test_dir.py
```diff
@@ -0,0 +1,53 @@
...
+
```
tshirtman:
```suggestion
```

tests/test_dir.py
```diff
@@ -0,0 +1,53 @@
...
+
```
tshirtman:
```suggestion
```

tshirtman:
I'm wondering now about the validity of replacing `__dir__` for java methods objects, wouldn't it be cleaner to have a specific attribute or method of the object, to call for instead of twisting the semantics of the dir() function for this object? https://docs.python.org/3/library/functions.html#dir 
> If the object has a method named `__dir__()`, this method will be called and must return the list of attributes. This allows objects that implement a custom `__getattr__()` or `__getattribute__()` function to customize the way dir() reports their attributes.

tests/test_dir.py
amsukdu: I thought matching exact strings are miscellaneous but it seems more appropriate than just plain comments.

amsukdu:
Yes, my first intention was piggybacking to `__dir__` method. The semantics doesn't perfectly match but I would say it is in the same ballpark with the original. The original `dir()` from `JavaMethod` was not helpful for me to understand the object. Many python users would poke around with `dir()` or `help()` so it could be a trade-off between strict policy and make another custom method, or a little bit of flexibility.

I'm totally okay with either way :)

tshirtman:
dir indeed is useful to look into the object, but adding a `signature` method will also be visible in the result of dir() and i've no doubt that it'll get the attention of anyone looking for that information.

amsukdu:
Agreed. Document this `signature` method to the `README` can be very helpful to the new comers.

tshirtman:
Closing this one as i merged #467 thanks a lot!

PR #429, opened 2019-07-01
Add build support for Oracle Solaris on SPARC sun4u and sun4v systems.

ggates1020:
Added configuration information to build on Oralce Solaris on SPARC sun4u and sun4v systems.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -195,7 +195,9 @@ else:
     MACHINE2CPU = {
         "i686": "i386",
         "x86_64": "amd64",
-        "armv7l": "arm"
+        "armv7l": "arm",
+        "sun4u": "sparcv9",
+        "sun4v": "sparcv9"
     }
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -210,6 +212,9 @@ def compile_native_invocation_handler(*possible_homes):
     if PLATFORM == 'win32':
         INCL_DIR = join(JDK_HOME, 'include', 'win32')
         LIBRARIES = ['jvm']
+    elif PLATFORM == 'sunos5':
+        INCL_DIR = join(jdk_home, 'include', 'solaris')
```
AndreMiras: I don't see where the lower-case `jdk_home` is defined? A linter would have caught that :cry:

AndreMiras: https://github.com/kivy/pyjnius/pull/446

PR #444, opened 2019-09-17
Use Github Actions for tests and releases

tshirtman:


AndreMiras:
Nice initiative! What about relying on our `Makefile` more so it's more portable across CI and environments?
What I mean is for instance instead of having the tests prep and run "hardcoded" in the `- name: test` section, we could use what's already available in the [`tests` target of our Makefile](https://github.com/kivy/pyjnius/blob/d5468fb/Makefile#L31).
Also it's a different topic, but I'm wondering why we don't use `tox` here to test against multiple targets? Maybe `tox` is too old or cumbersome of pyjnius, but I'm wondering

tshirtman:
@AndreMiras i think i tried using the makefile at first but switched to hard coding because something was going wrong, i can try going that way again now things seem to work well, about tox, i actually never used that, maybe it could simplify things, i don't know.

I did left python2.7 aside (and any python before 3.6 actually), except on osx where i didn't research yet how to test more recent versions, i guess there should be ways to use brew to install them.

tshirtman:
I think this is ready to merge, review welcome :)

PR #450, opened 2019-09-24
Do on better way Space validation

jrnp97:
Pull request referent to comment on thread https://github.com/kivy/pyjnius/pull/420#discussion_r327866825


tshirtman:
Thanks for the PR, but the code changed quite a bit anyway when moving to github actions, as i needed to change a lot of things in setup.py to make wheels, long story short, this change has been obsoleted, sorry about that.

PR #459, opened 2019-12-03
make sure Interface have access to Object's methods

### LINKED ISSUE #455: cast to interface, java.lang.Object methods not available
jnius doesnt expose methods of java.lang.Object for references cast to interfaces.

For instance, the following Java code works:
```
import java.util.Map
import java.util.HashMap
var hmap = new HashMap<String,String>();
hmap.put("a", "1")
((Map<String,String>)hmap).toString()
```

The following Python does not:
```
from jnius import autoclass
from jnius import cast
mapClass = autoclass('java.util.HashMap')
hmap = mapClass()
hmap.put("a", "1")
print(thehmap.toString())
mmap = cast('java.util.Map', hmap)
mmap.toString()
```
gives
> AttributeError: 'java.util.Map' object has no attribute 'toString'

I would contend that any instantiated interface must extend Object, by virtue of the fact that it is instantiated.

Similar problems with .getClass()

tshirtman:
as specified in https://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html#jls-9.2

fixes #455

### DIFF
--- a/jnius/jnius_export_func.pxi
+++ b/jnius/jnius_export_func.pxi
@@ -1,5 +1,6 @@
 from cpython.version cimport PY_MAJOR_VERSION
 
+
 def cast(destclass, obj):
     cdef JavaClass jc
     cdef JavaClass jobj = obj
@@ -11,6 +12,7 @@ def cast(destclass, obj):
         jc = destclass(noinstance=True)
     jc.instanciate_from(jobj.j_self)
     return jc
+
 
 def find_javaclass(namestr):
     namestr = namestr.replace('.', '/')
--- a/jnius/reflect.py
+++ b/jnius/reflect.py
@@ -11,7 +11,7 @@ from .jnius import (
     JavaException
 )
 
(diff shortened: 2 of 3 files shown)


cmacdonald:
i gave a browse. As the test case (test_mmap_toString) passes, its good. I was wondering if you expected any significant overhead in the additional logging?

cmacdonald:
Is it worth adding
mmap.getClass() to the test case also?
My original issue reported getClass() being an issue, but from the code you didn't need to fix anything for getClass()

tshirtman:
Yeah, nothing specific to getClass, (and i don't even think the methods declaration i added on Object are needed, will check that), i can add it to the test, but i'm not sure it's needed.
I didn't compare the performance of the logging, but as by default the level is not set, these debug prints won't be made, i don't think it should be really costly, and it's only when the class is constructed, so it's a one time cost, not every time say, a method is called, so i think the cost outweighs the benefit.

PR #460, opened 2019-12-05
Keep quiet when deducing jdk_home from JAVA_HOME

wdscxsj:
It's standard practice of Java to set the environment variable `JAVA_HOME` to the JDK installation directory. Thus it's no surprise to find a `bin/javac` under it. Better print nothing (esp. directly to stdout) when the detection is successful.

By the way, I don't think `JDK_HOME` is a good environment variable name for `$JAVA_HOME/bin`. None of the conventional names like `R_HOME` , `GOROOT`, `GOPATH` etc. points to a `bin` directory.

### DIFF
--- a/jnius/env.py
+++ b/jnius/env.py
@@ -117,7 +117,6 @@ def get_jdk_home(platform):
     if not jdk_home:
         if platform == 'win32':
             TMP_JDK_HOME = getenv('JAVA_HOME')
-            print(TMP_JDK_HOME)
             if not TMP_JDK_HOME:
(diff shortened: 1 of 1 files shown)


tshirtman:
Thanks, yeah, i'm also unsure about the JDK_HOME variable, it was already there before the refactoring i did, i tried to sanitize the logic, but there is certainly room for improvement still, maybe it's better to eliminate it completely, and instead rely either on finding javac in PATH or looking for it in JAVA_HOME whatever the case.

Issue #405, opened 2019-02-10
Fix ' TypeError: 'float' object is not callable '


dschiller:
Found a Solution for the '# pat.similar(1.0) # Throw an Error: TypeError: 'float' object is not callable':

In 'C:\Users\DOP\AppData\Local\Programs\Python\Python37\Lib\site-packages\jnius\reflect.py' change ( Line 188 with pyjnius Version 1.2.0:

            if name != 'getClass' and bean_getter(name) and len(method.getParameterTypes()) == 0:
                lowername = lower_name(name[2 if name.startswith('is') else 3:])
                classDict[lowername] = (lambda n: property(lambda self: getattr(self, n)()))(name)
            continue

to:

            if name != 'getClass' and bean_getter(name) and len(method.getParameterTypes()) == 0:
                lowername = lower_name(name[2 if name.startswith('is') else 3:])
                if lowername not in methods_name:
                 classDict[lowername] = (lambda n: property(lambda self: getattr(self, n)()))(name)
            continue

suhassumukhv:
@dschiller Thank you. This is indeed a bug. I can second that your solution resolves the issue.

dschiller:
Cool. But it is a fix from @jianbing ( https://github.com/jianbing/sikulitool/issues/1#issuecomment-327686662 )

tshirtman:
Can you do a PR with it? Will try to check asap.

tshirtman:
should be fixed by #518

PR #464, opened 2019-12-07
Avoid replacing existing fields with getter methods

tshirtman:
from @dschiller & @jianbing
https://github.com/kivy/pyjnius/issues/405


cmacdonald:
this can now be closed?

PR #466, opened 2019-12-14
Issue 465

tshirtman:


jnius/reflect.py
```diff
@@ -219,7 +219,10 @@ def autoclass(clsname):
 
     cls = c
     while cls is not None:
-        methods = cls.getDeclaredMethods()
+        if cls is c:
+            methods = cls.getDeclaredMethods()
+        else:
```
cmacdonald: suggest that you explicitly state in comment for else branch that cls is a superclass of c, as otherwise we have to read to end of code.

PR #489, opened 2020-03-02
improvements to specified constructor signature selection

cmacdonald:
this:
 - fixes an issue where the requested signature is not the first one found
 - forces an exception when a signature is requested, but one was not found


tshirtman:
Not sure why there is no CI running on that, i guess we'll merge and see if we have things to fix, better than let fixes rot.

PR #495, opened 2020-04-03
just build for java10 for testing purpose

tshirtman:


tshirtman:
ok, so java10 and python3.6 are not friends in our tests (segault), except on osx, where it works, i didn't try looking into the origin of the segfault, mainly because i don't have a way to reproduce localy (maybe i'll try to find some java 10 docker image and build pyjnius in it), but at least it seems to be limited to this version combination.

Issue #494, opened 2020-04-02
Add __exit__ for java.io.Closable or java.io.Autoclosable methods


cmacdonald:
Patch for autoclass. For bonus points, it would be good to have an API where we could automatically decorate newly created object with appropriate dunder methods.

cmacdonald:
#497 merged. Thank you @tshirtman. Closing.

tshirtman:
no thank *you* :), the way you implemented it is great, as it'll allow making using java objects in a lot of more pythonic ways (support for iter(), `in`, and probably others) <3

PR #497, opened 2020-04-04
add support for java.lang.AutoCloseable #494

cmacdonald:
Here is a patch for #494
Some possible areas for feedback: 
 - is calling find_javaclass() for every interface in interface_map everytime autoclass() is called expensive?
 - isAssignableFrom() doesnt just check interfaces, we can also check all classes. So perhaps interface_map isnt the best name. 
 - Should there be a method for accessing interface_map instead of exporting it directly?
 - what happens if interface_map contains the class that is being called, e.g. 
```
from jnius import autoclass, interface_map
interface_map["x.y"] = { "__size__" : lambda self : self.size() }
autoclass("x.y")
```

### DIFF
--- a/jnius/reflect.py
+++ b/jnius/reflect.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import
 from __future__ import unicode_literals
 from __future__ import division
-__all__ = ('autoclass', 'ensureclass', 'interface_map')
+__all__ = ('autoclass', 'ensureclass', 'protocol_map')
 from six import with_metaclass
 import logging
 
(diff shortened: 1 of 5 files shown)


tshirtman:
i don't necessarily have answers for all these questions, but i can try :)
- i would assume not, i do assume most autoclass intense usage is at the start of an app, and if there is any autoclass usage during the lifetime, it's going to be very sparse, so i don't think we have to be exceptionally careful about performances here, and i don't think find_javaclass is slow enough to justify particular caution, but if in doubt, we can mesure performances.
- good question, i think it's more interesting to look at the python side of things, this mapping of "dunder" methods/attributes to features is called a `protocol` in the python world, so i would call this `protocol_map` rather.
- in python, i think the best practice is to assume people are grown ups and they are able to update a dictionary carefuly enough not to break what other parts of the code could have put here, of course we could provide an interface around it, but in the end, we are going to need to provide almost the same interface as a dictionary, or prevent some legitimate use cases, so i'd be ok with just keeping this as a simple exposed attribute, if problems arise later, we can always make that a property with getter and setters, as is commonly done in python, to add sanity checks.
- i don't really see how that's a problem? the code is going to be called after instantiation anyway?

cmacdonald:
OK. So, given that #501 has landed, I could rewrite this to address questions 1, 2 & 4. I renamed it to protocol_map. Given your feedback, I suggest we don't add an API for now for question 3.

cmacdonald:
I think this is good for re-review.

jnius/reflect.py
```diff
@@ -335,3 +323,31 @@ def _getitem(self, index):
...
+            # python for...in iteration checks for end of list by waiting for IndexError
+            raise IndexError()
+        else:
+            raise
+
+# protocol_map is a user-accessible API for patching class instances with additional methods 
+protocol_map = {
+    'java.util.List' : {
+        '__getitem__' : _getitem,
+        '__len__' : lambda self: self.size()
```
tshirtman, line 346: I think `__len__` -> .size() can actually be applied to the super interface `java.util.Collection`.

tests/test_closeable.py
```diff
@@ -0,0 +1,38 @@
...
+import unittest
+from jnius import autoclass, JavaException, protocol_map
+
+
+class TestConstructor(unittest.TestCase):
+    '''
+    TestCase for using java.io.Closeable dunder
+    '''
+
+    # this checkes that java.lang.AutoCloseable instances gain the correct dunder methods
```
tshirtman, line 15: I believe these comments can be made into docstrings of the test. Anyway they should be inside the method, not before (same for next test).

tshirtman:
Possible… damn phone app won't let me correct. :(

cmacdonald:
addressed feedback, and ensured to test len() in various places

Issue #465, opened 2019-12-13
superclass methods not accessible in 1.2.1


cmacdonald:
```python
arraylist = autoclass("java.util.ArrayList")()
arraylist.iterator()
arraylist.stream()
```

This works for 1.2.0 but not for 1.2.1.
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-7-5e67e1c90388> in <module>()
----> 1 arraylist.stream()

AttributeError: 'java.util.ArrayList' object has no attribute 'stream'
```

Reproducible notebook at https://colab.research.google.com/drive/1F9u2jVQR5JFw_mk5Bq--VH1Ki91Xe5x3

stream() is defined in the super-interface as a default.

We also had problem accessing methods in interfaces that extended java.util.List.

tshirtman:
I can reproduce it locally, let's see if it also happens in CI, i'll look for a fix when i have some time.

cmacdonald:
I can't see how we broke it. I can see that getDeclaredMethods() doesn't include methods that are default implementations.  (They arent declared in the class or its supertypes).

There is some discussion in https://blog.jooq.org/2018/03/28/correct-reflective-access-to-interface-default-methods-in-java-8-9-10/ - I dont _think_ this is relevant.

Is the solution also to walk all interfaces and their parent interfaces adding default methods? (We could add the abstract methods, but this seems superfluous as they must by implemented by the concrete object).

This is the most relevant stackoverflow thread I found: https://stackoverflow.com/questions/28400408/what-is-the-new-way-of-getting-all-methods-of-a-class-including-inherited-defau

it does seems that there could be some differences in different JVM implementations...

tshirtman:
It does work if i cast to java.util.Collection first, but it shouldn't be necessary… :/

tshirtman:
I also think it's because of using getDeclaredMethods instead of getMethods, i tried to walk all the interfaces but somehow it wasn't finding the Collection interface in getInterfaces()…

tshirtman:
hm, this seems to work… 
https://github.com/kivy/pyjnius/pull/466/files

tshirtman:
https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=100815&view=logs&j=696704cc-6fef-57a3-ea36-f27779b8cd5e&t=06421391-4b55-523d-a804-5e1a5bfc1908
[...]

PR #500, opened 2020-04-07
add include_protected and include_private options to autoclass

hx2A:
Adding a `public_only` option to the `autoclass` function. When used this can limit the created class to only public methods and fields.

### DIFF
--- a/jnius/reflect.py
+++ b/jnius/reflect.py
@@ -196,7 +196,7 @@ def log_method(method, name, signature):
     )
 
 
-def autoclass(clsname):
+def autoclass(clsname, public_only=False):
     jniname = clsname.replace('.', '/')
     cls = MetaJavaClass.get_javaclass(jniname)
     if cls:
@@ -222,9 +222,9 @@ def autoclass(clsname):
     while cls is not None:
         level += 1
         if cls is c:
-          methods = cls.getDeclaredMethods()
+            methods = cls.getDeclaredMethods()
         else:
-          methods = cls.getMethods()
+            methods = cls.getMethods()
         methods_name = [x.getName() for x in methods]
 
         for index, method in enumerate(methods):
@@ -235,6 +235,8 @@ def autoclass(clsname):
             # only one method available
             if methods_name.count(name) == 1:
                 static = Modifier.isStatic(method.getModifiers())
+                if public_only and not Modifier.isPublic(method.getModifiers()):
+                    continue
                 varargs = method.isVarArgs()
                 sig = '({0}){1}'.format(
                     ''.join([get_signature(x) for x in method.getParameterTypes()]),
@@ -253,6 +255,8 @@ def autoclass(clsname):
                 if subname != name:
                     continue
                 method = methods[index]
+                if public_only and not Modifier.isPublic(method.getModifiers()):
+                    continue
                 sig = '({0}){1}'.format(
                     ''.join([get_signature(x) for x in method.getParameterTypes()]),
                     get_signature(method.getReturnType()))
@@ -290,6 +294,8 @@ def autoclass(clsname):
 
     for field in c.getFields():
         static = Modifier.isStatic(field.getModifiers())
+        if public_only and not Modifier.isPublic(field.getModifiers()):
+            continue
         sig = get_signature(field.getType())
         cls = JavaStaticField if static else JavaField
         classDict[field.getName()] = cls(sig)
--- a/jnius/reflect.py
+++ b/jnius/reflect.py
@@ -294,7 +294,7 @@ def autoclass(clsname, public_only=False):
             classDict['__len__'] = lambda self: self.size()
             break
 
-    for field in c.getFields():
+    for field in c.getDeclaredFields():
         field_modifiers = field.getModifiers()
         static = Modifier.isStatic(field_modifiers)
         if public_only and not Modifier.isPublic(field_modifiers):
--- a/jnius/reflect.py
+++ b/jnius/reflect.py
@@ -234,8 +234,9 @@ def autoclass(clsname, public_only=False):
 
             # only one method available
             if methods_name.count(name) == 1:
-                static = Modifier.isStatic(method.getModifiers())
-                if public_only and not Modifier.isPublic(method.getModifiers()):
+                method_modifiers = method.getModifiers()
+                static = Modifier.isStatic(method_modifiers)
+                if public_only and not Modifier.isPublic(method_modifiers):
                     continue
                 varargs = method.isVarArgs()
                 sig = '({0}){1}'.format(
@@ -255,7 +256,8 @@ def autoclass(clsname, public_only=False):
                 if subname != name:
                     continue
                 method = methods[index]
-                if public_only and not Modifier.isPublic(method.getModifiers()):
+                method_modifiers = method.getModifiers()
+                if public_only and not Modifier.isPublic(method_modifiers):
                     continue
                 sig = '({0}){1}'.format(
                     ''.join([get_signature(x) for x in method.getParameterTypes()]),
@@ -263,7 +265,7 @@ def autoclass(clsname, public_only=False):
 
                 if log.level <= logging.DEBUG:
                     log_method(method, name, sig)
-                signatures.append((sig, Modifier.isStatic(method.getModifiers()), method.isVarArgs()))
+                signatures.append((sig, Modifier.isStatic(method_modifiers), method.isVarArgs()))
 
             classDict[name] = JavaMultipleMethod(signatures)
 
@@ -293,8 +295,9 @@ def autoclass(clsname, public_only=False):
             break
 
     for field in c.getFields():
-        static = Modifier.isStatic(field.getModifiers())
-        if public_only and not Modifier.isPublic(field.getModifiers()):
+        field_modifiers = field.getModifiers()
+        static = Modifier.isStatic(field_modifiers)
+        if public_only and not Modifier.isPublic(field_modifiers):
             continue
         sig = get_signature(field.getType())
         cls = JavaStaticField if static else JavaField
--- a/jnius/reflect.py
+++ b/jnius/reflect.py
@@ -211,7 +211,7 @@ def identify_hierarchy(cls, level, concrete=True):
     yield cls, level
 
 
-def autoclass(clsname, include_protected=True, include_private=True):
+def autoclass(clsname, include_protected=False, include_private=False):
     jniname = clsname.replace('.', '/')
     cls = MetaJavaClass.get_javaclass(jniname)
     if cls:
@@ -292,7 +292,7 @@ def autoclass(clsname, include_protected=True, include_private=True):
             # multiple signatures
             signatures = []
             log.debug("method %s has %d multiple signatures in hierarchy of cls %s" % (name, len(cls_methods[name]), c))
-            
+
             paramsig_to_level=defaultdict(lambda: float('inf'))
             # we now identify if any have the same signature, as we will call the _lowest_ in the hierarchy,
             # as reflected in min level
@@ -325,13 +325,6 @@ def autoclass(clsname, include_protected=True, include_private=True):
         if cls_name in protocol_map:
             for pname, plambda in protocol_map[cls_name].items():
                 classDict[pname] = plambda  
-
-    # for field in c.getFields():
-    #     print(f'adding field {field.getName()}')
-    #     static = Modifier.isStatic(field.getModifiers())
-    #     sig = get_signature(field.getType())
-    #     cls = JavaStaticField if static else JavaField
-    #     classDict[field.getName()] = cls(sig)
 
     for field_name, (field, _) in cls_fields.items():
         field_modifier = field.getModifiers()
--- a/jnius/reflect.py
+++ b/jnius/reflect.py
@@ -211,9 +211,12 @@ def identify_hierarchy(cls, level, concrete=True):
     yield cls, level
 
 
+# NOTE: if you change the include_protected or include_private default values,
+# you also must change the classparams default value in MetaJavaClass.__new__
+# and MetaJavaClass.get_javaclass.
 def autoclass(clsname, include_protected=False, include_private=False):
     jniname = clsname.replace('.', '/')
-    cls = MetaJavaClass.get_javaclass(jniname)
+    cls = MetaJavaClass.get_javaclass(jniname, classparams=(include_protected, include_private))
     if cls:
         return cls
 
@@ -342,7 +345,8 @@ def autoclass(clsname, include_protected=False, include_private=False):
         MetaJavaClass,
         clsname,
         (JavaClass, ),
-        classDict)
+        classDict,
+        classparams=(include_protected, include_private))
 
 
 ## dunder method for List
--- a/jnius/reflect.py
+++ b/jnius/reflect.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import
 from __future__ import unicode_literals
 from __future__ import division
-__all__ = ('autoclass', 'ensureclass')
+__all__ = ('autoclass', 'ensureclass', 'protocol_map')
 from six import with_metaclass
 import logging
 
@@ -12,6 +12,7 @@ from .jnius import (
 )
 
 log = logging.getLogger(__name__)
+from collections import defaultdict
 
 
 class Class(with_metaclass(MetaJavaClass, JavaClass)):
@@ -48,7 +49,7 @@ class Class(with_metaclass(MetaJavaClass, JavaClass)):
     getSigners = JavaMethod('()[Ljava/lang/Object;')
     getSuperclass = JavaMethod('()Ljava/lang/Class;')
     isArray = JavaMethod('()Z')
-    isAssignableFrom = JavaMethod('(Ljava/lang/reflect/Class;)Z')
+    isAssignableFrom = JavaMethod('(Ljava/lang/Class;)Z')
     isInstance = JavaMethod('(Ljava/lang/Object;)Z')
     isInterface = JavaMethod('()Z')
     isPrimitive = JavaMethod('()Z')
@@ -195,8 +196,22 @@ def log_method(method, name, signature):
         Modifier.isStrict(mods)
     )
 
+def identify_hierarchy(cls, level, concrete=True):
+    supercls = cls.getSuperclass()
+    if supercls is not None:
+         for sup, lvl in identify_hierarchy(supercls, level + 1, concrete=concrete):
+             yield sup, lvl # we could use yield from when we drop python2
+    interfaces = cls.getInterfaces()
+    for interface in interfaces or []:
+        for sup, lvl in identify_hierarchy(interface, level + 1, concrete=concrete):
+            yield sup, lvl
+    # all object extends Object, so if this top interface in a hierarchy, yield Object
+    if not concrete and cls.isInterface() and not interfaces:
+        yield find_javaclass('java.lang.Object'), level +1
+    yield cls, level
 
-def autoclass(clsname, public_only=False):
+
+def autoclass(clsname, include_protected=True, include_private=True):
     jniname = clsname.replace('.', '/')
     cls = MetaJavaClass.get_javaclass(jniname)
     if cls:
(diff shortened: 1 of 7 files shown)


tshirtman:
looks good, but:
- would be nice to have some tests :)
- if we can avoid calling getModifiers multiple times on the same object, that would be nice, -> save the result to a "modifiers" local variable.

hx2A:
I added code to save the result of getModifiers to a local variable.

I'm having difficulty running the current unit tests. Why does it keep trying to use Python 2.7?

The current tests folder has some java code but it looks like every class has only public methods. How's this for a unit test: I'll create a new class with public and private methods and fields and use `autoclass` twice, once with `public_only=True`. Then I'll verify that the private methods and fields are not in both classes.

tshirtman:
That seems like a good test, yes :).

For the tests, make sure you have pytest installed for python3 (or that you are in a python3 virtualenv, with pytest installed).

cmacdonald:
hi guys, two comments. 

1. It would make sense for public_only=true to be default. I would reverse the logic to be `include_private`, and set to False by default. That would break backwards functionality, but for in the unlikely case that anyone is relying on a private Java API being accessible from Python.

2. its likely to conflict with PR #501. But I guess we can sort that out later.

tshirtman:
1. yes, it would make sense, but that's kind of an api break, and it might surprise users (yes we did surprise them already by breaking things, but that wasn't intentional :])
2. yes, but that shouldn't be too hard to fix indeed.

cmacdonald:
Proposa:
 - default in this series: private=None; and if so, emit a log warning in JavaMethod.__call__(), but the first time that method is called
 - if private = False, dont include the private methods
 - if private = True,  include the private methods without the warning

tshirtman:
about the name, now that i think about it, include_private is more flexible of a name than public_only (e.g if later, someone wants a flag to include protected/native/abstrat or some other filter, public_only would prevent us from having that). 
about the warning, you mean the first time autoclass is called, or the first time each private method is called? The second seems a bit noisy, but if it's easy to disable by passing the include_private=True to autoclass, maybe that's fine.

hx2A:
> about the name, now that i think about it, include_private is more flexible of a name than public_only (e.g if later, someone wants a flag to include protected/native/abstrat or some other filter, public_only would prevent us from having that).

That's a good point, a more flexible name is the better choice. I'll change it to be "include_private" instead of "public_only".

For my particular use case, I want only public methods, which is why I made the pull request. Is it OK with you if I add a "include_protected" option also?

> about the warning, you mean the first time autoclass is called, or the first time each private method is called? The second seems a bit noisy, but if it's easy to disable by passing the include_private=True to autoclass, maybe that's fine.

I might not be the best person to code those warnings as I am new to this project, but am happy to help out if I get some guidance.

tshirtman:
sure, adding include_protected should be fine too.

hx2A:
I started creating unit tests and have run into two problems.

1. In the beginning of the `autoclass` function it checks `MetaJavaClass` to see if it has created a class with that name before. If so, it retrieves it and returns it to the caller.

```
    jniname = clsname.replace('.', '/')
    cls = MetaJavaClass.get_javaclass(jniname)
    if cls:
        return cls
```

This makes it difficult to create proper unit tests because I need to create multiple versions of the same class with different visibility levels set. I suppose I could add another parameter like `check_class_cache` but I suspect this could or would create other problems.

There are two new unit test files: `test_visibility_all.py` and `test_visibility_public_only.py`. Each will run correctly if the other is commented out.

This isn't just a unit test issue: it is possible a user might want to create multiple versions of the same class with different method visibilities. Whichever version is created first will be the one returned for all the other calls, leading to a confusing bug.

2. It seems that the current `autoclass` function will create a class with public, protected, and private _methods_ but only public _fields_. Unless if I'm doing something wrong with my tests or Java code. I have pushed the current state of the unit tests so you can see what I am doing. I'm not sure what's going on here but it looks like in `autoclass` the loop `for field in c.getFields()` only loops over the public fields.

It seems to me it would be confusing to add parameters for including/excluding private and protected methods but have it not work the same way for fields.

Thoughts? I have some ideas but I am not sure how to proceed.

hx2A:
After sleeping on this, I'd like to propose a new direction here. How about we scrap this pull request and do this instead:

1. Enter a bug for `autoclass` and how it creates a class with public, protected, and private methods but only public fields. It should either be consistent and provide protected and private fields or the `autoclass` documentation should be updated to let people know this is how it works.

2. Update the documentation to let people know that `autoclass` will give you a class with protected and private methods. This is definitely how the function should work, and if people want something customized they can use the API or modify the output of `autoclass`. (It seems I can just `del` functions from the returned class to remove methods from instances.)

3. Update the documentation to include the importable function `find_javaclass` and how it might be used to facilitate the creation of classes with the API. Speaking as a pyjnius noob, it seems like there is a big knowledge gap between successfully using `autoclass` and successfully using the API for a non-trivial implementation. It would be helpful to fill in that gap.

And I'm happy to write the documentation, with feedback from others. Given that I am new to this library, my perspective on what information is useful would help other new people.

tshirtman:
I'm curious about what changed your mind on 2. I mean, i agree the inconsistency between methods and fields is a problem, but i don't follow how/if that impacts the decision to allow or not people to select what they want.
If you want to be able to select what you want, then we'll want to add the modifiers data as metadata to the methods and fields, and i'm not sure how to do that in a nice way, creating the data is easy, but how to expose it?
- As attributes of each field/method? there is a risk of overriding some existing attribute, or the names will be so mangled that they are going to be inconvenient to use.
- As attributes of the class? a dictionary of sets/lists for private/public/protected/etc attributes so the user can check if any attribute they are accessing is in the set that they care about? while correct, that seems cumbersome to use.
- Any other idea?

tshirtman:
> This isn't just a unit test issue: it is possible a user might want to create multiple versions of the same class with different method visibilities. Whichever version is created first will be the one returned for all the other calls, leading to a confusing bug.
I'm not a java expert, but i have the feeling this would be frowned upon, i do think classes are supposed to have a unique fully qualified name, and importing it should always give you the same result.
To implement unittests, i would simply create different classes, with similar content, and compare the results produced by autoclass, on them, no need for them to have the exact same name, unless i'm missing something.
In fact, you could use the same class, but import it with autoclass multiple times, with different visibility options (include_private=True or False), it would indeed show an issue with caching in autoclass, and the correct way to solve it would be to ensure the cache takes these flags into account, not just the class name.

hx2A:
> I'm curious about what changed your mind on 2. I mean, i agree the inconsistency between methods and fields is a problem, but i don't follow how/if that impacts the decision to allow or not people to select what they want.

Your opinion on this is more relevant than mine becuase you are a member of the Kivy team. I agree that people should be able to create a class as they want it to be. I also think that if the `autoclass` function is going to implement a feature, it should be done correctly or not at all. It seems like adding this feature to `autoclass` might not always work correctly or as people might expect.

People can always use the API if `autoclass` doesn't do what they want it to do. The `autoclass` function doesn't have to address everyone's needs. Right?

> If you want to be able to select what you want, then we'll want to add the modifiers data as metadata to the methods and fields, and i'm not sure how to do that in a nice way, creating the data is easy, but how to expose it?

I liked the idea of adding the modifiers as metadata to the class or the methods/fields, and I actually looked for this as I was learning about the library, thinking it would be available somewhere.

There is one complication with adding metadata though: what about JavaMultiMethods? It is possible for a Java class to have multiple methods with the same name but different params and modifiers.

> I'm not a java expert, but i have the feeling this would be frowned upon, i do think classes are supposed to have a unique fully qualified name, and importing it should always give you the same result.

Good point, it would be frowned upon.

> To implement unittests, i would simply create different classes, with similar content, and compare the results produced by autoclass, on them, no need for them to have the exact same name, unless i'm missing something.

Right, I can do that. Not sure why I didn't think of that.

tshirtman:
> There is one complication with adding metadata though: what about JavaMultiMethods? It is possible for a Java class to have multiple methods with the same name but different params and modifiers.

Good point. As the JavaMultipleMethod is what gets exposed to the user, they can't check if what's going to be called (depending on the parameters they give), is private or not, maybe we could imagine adding methods to the method (so you'd do `object.methodname.public()(arguments)` instead of `object.methodname(arguments)` but i don't like it much, i think using autoclass to get different versions of the class is cleaner, i guess you could even cast an object that was created with the "private" version of the class to a "public" version of the class, and then not be able to call the methods/fieds that aren't public, (and vice-versa, using the cast to the private version would make them visible again). To achieve that, i think on top of what you actually did before, the cache system has to be changed to use the visibility parameters (private, protected) as well as the name of the class. I think that should be enough to do it, but we certainly should fix the issue with the private/protected fields not being exposed first.

tshirtman:
BTW, i don't use pyjnius often these days, so my opinion is not necessarily the most relevant, i just care about it because it's an important project for kivy and it has been receiving far less attention than it needs, but actual users of the projects absolutely have their say in how to go forward with it, and improvements to it are very welcome, assuming we don't break the current important usages of it on android.

cmacdonald:
Here's a different take. Thinking about the Java security model:
 - protected methods/fields are only available to sub-classes. Pyjnius cant create sub-classes, so they shouldn't be accessible at all.
 - private methods/fields are only accessible to the class itself
 - package protected (i.e. no declaration) are only accessible in the same package. We may simply assume that python classes aren't in the same package.

I would contend that public_only _should_ be the default that is accessible to Python (previous version aside). I  use Pyjnius to make a library accessible to Python - I cant see any use case to make private/protected methods/fields available (though I note a request on https://github.com/kivy/pyjnius/issues/465#issuecomment-611081006 that requests access to a protected field...)

hx2A:
> To achieve that, i think on top of what you actually did before, the cache system has to be changed to use the visibility parameters (private, protected) as well as the name of the class. I think that should be enough to do it, 

Cool, I will take care of this.

> but we certainly should fix the issue with the private/protected fields not being exposed first.

Fixed!

Referring to the [java.lang.Class](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html) documentation, I saw that [getFields](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getFields--) was being used, which only returns the public fields. Another method, [getDeclaredFields](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getDeclaredFields--), would give all the fields. Changing that fixed the problem.

hx2A:
> Fixed!

Not fixed. My fix created a problem with inherited fields. I'll get it working again.

tshirtman:
> > Fixed!
> 
> Not fixed. My fix created a problem with inherited fields. I'll get it working again.

yeah, i think you re going down the same path i went before, look at the fix in #501 i think applying the same lookup order to build the list of fields should fix it as well.

tshirtman:
> protected methods/fields are only available to sub-classes. Pyjnius cant create sub-classes, so they shouldn't be accessible at all.

That's a sad technical limitation that we didn't manage to work around, i'd love for this to be possible, and i tried very hard and couldn't make it work. In the mean time, i'd rather let the users use any mean at their disposal to implement the features they want to in python rather than in ava.

> I would contend that public_only _should_ be the default that is accessible to Python (previous version aside). I use Pyjnius to make a library accessible to Python - I cant see any use case to make private/protected methods/fields available (though I note a request on [#465 (comment)](https://github.com/kivy/pyjnius/issues/465#issuecomment-611081006) that requests access to a protected field...)

from a correctness point of view, i agree, from a practical one, i'm not sure, as pyjnius is first and foremost there to allow doing things that are normally only possible in java on android, i'd rather not remove working features, just to be more correct regarding regarding theorical properties of the java VM (as afaik it's always possible to bypass private/protected status, it's just a matter of how ugly you want the code to be, the fact that we are able to do it in pyjnius is proof it's not a foolproof feature), in python privacy is mostly a declaration of intent, the mesures to enforce it are very light. I think it's fine to ask pyjnius not to present us the things we are not supposed to see, be it opt in or opt-out, either is fine, it's just that if anybody depends on code that used private/protected attributes/methods right now, i'd rather avoid break their code until at least a 2.0 (i think first place to check would be plyer, but surely there are a lot of other usages of pyjnius in the wild, considering our almost 1k stars and 200+ forks). So for now i'd prefer if the option would be added as opt-out (of seeing the private things).

cmacdonald:
Sure @tshirtman, thats fine with me.

hx2A:
It looks like @cmacdonald has made some changes to `reflect.py` and `autoclass` in particular, so I'm no longer working with the most up to date version.

One of the problems I identified last night has to do with functions with the same name at multiple levels of an inheritance hierarchy and how all of them needed to end up in the `JavaMultiMethod`. Previously only some of them would be used. @cmacdonald , do your changes address this? It looks to me as if it does.

The other problem is for Interfaces with fields. Interface fields are always `public static final` and are perhaps not a recommended coding practice, but my specific use case has an Interface with a lot of them so I'd like it if `autoclass` supported this also.

I'll continue working using the newer version of `reflect.py` and add the `include_protected` and `include_private` params and address the previously mentioned issues with fields.

cmacdonald:
> One of the problems I identified last night has to do with functions with the same name at multiple levels of an inheritance hierarchy and how all of them needed to end up in the JavaMultiMethod. Previously only some of them would be used. @cmacdonald , do your changes address this? It looks to me as if it does.

Yes, I think this is now solved in the master branch through #501.

I can give some input with this. I understand that child fields override super fields. I think a solution would look like: `cls_fields={}`

then inside loop for cls,level in class_hierachy:
```python
fields = cls.getDeclaredFields()
for field in fields:
  field_name = field.getName()
  if field_name in cls_fields:
    if level < cls_fields[field_name](1):
      cls_fields[field.getName] = (field, level)
  else:
    cls_fields[field.getName] = (field, level)
```

After that, you would have to walk cls_fields, adding them as appropriate.

hx2A:
I just checked in a version that seems to implement the `include_protected` and `include_private` params correctly. I need to create more unit tests to verify this, but I think it is correct.

The default values for both is `False`. It turns out this is necessary to make the other unit tests work. The [java.util.ArrayList](https://docs.oracle.com/javase/8/docs/api/java/util/ArrayList.html) class has a method `size()` and a private `int size`. If the default value is `True`, calling `autoclass` creates a class with a `size` attribute that is an `int`, not a callable `JavaMultipleMethod`.

This code takes care of what I said before about static fields in an Interface. I didn't have to add anything special to make that work.

There is one bug that gets on my nerves, and I verified this has nothing to do with `autoclass` and exists in the current release of pyjnius: If I create a class with static fields and instantiate an instance of the class, if I try to access a static field from the instance, the JVM segfaults. This is particularly annoying when working with objects in IPython and I use the tab complete functionality.

Next steps:

* I need to make the `MetaJavaClass` class cache take into account the `include_protected` and `include_private` settings.
* More thorough unit tests

hx2A:
In the core dump error file I see this:

```
Internal exceptions (2 events):
Event: 0.025 Thread 0x0000561f72198000 Exception <a 'java/lang/NoSuchMethodError': Method sun.misc.Unsafe.defineClass(Ljava/lang/String;[BII)Ljava/lang/Class; name or signature does not match> (0x000000076ab07cc0) thrown at [/HUDSON3/workspace/8-2-build-linux-amd64/jdk8u74/6087/hotspot/src/sha
Event: 0.025 Thread 0x0000561f72198000 Exception <a 'java/lang/NoSuchMethodError': Method sun.misc.Unsafe.prefetchRead(Ljava/lang/Object;J)V name or signature does not match> (0x000000076ab07fa8) thrown at [/HUDSON3/workspace/8-2-build-linux-amd64/jdk8u74/6087/hotspot/src/share/vm/prims/jni.cp
```

cmacdonald:
good spot on the segfault. I can reproduce. I will give a look to see if I can figure it out.

Viz size, I wonder if we should prefer methods over variables, as that retain more of the jnius current behaviour, by permitting include_protected and include_private to default to True.

hx2A:
> I will give a look to see if I can figure it out.

Thanks, it would be fantastic if you could fix that.

> Viz size, I wonder if we should prefer methods over variables, as that retain more of the jnius current behaviour, by permitting include_protected and include_private to default to True.

I would be in favor of this. It's also an easy change to make.

tshirtman:
> The default values for both is `False`. It turns out this is necessary to make the other unit tests work. The [java.util.ArrayList](https://docs.oracle.com/javase/8/docs/api/java/util/ArrayList.html) class has a method `size()` and a private `int size`. If the default value is `True`, calling `autoclass` creates a class with a `size` attribute that is an `int`, not a callable `JavaMultipleMethod`.

:man_facepalming:  damn, that's a "fun" one, i guess it makes sense to default to False then, i don't know how many people currently depends on private methods, but maybe we should bump the version to 2.0 after merging that then. Considering the significant recent improvements, that doesn't seem unwarranted anyway.

> There is one bug that gets on my nerves, and I verified this has nothing to do with `autoclass` and exists in the current release of pyjnius: If I create a class with static fields and instantiate an instance of the class, if I try to access a static field from the instance, the JVM segfaults. This is particularly annoying when working with objects in IPython and I use the tab complete functionality.

Ouch indeed.

> 
> Next steps:
> 
>     * I need to make the `MetaJavaClass` class cache take into account the `include_protected` and `include_private` settings.
> 
>     * More thorough unit tests

:+1:

tshirtman:
> > I will give a look to see if I can figure it out.
> 
> Thanks, it would be fantastic if you could fix that.
> 
> > Viz size, I wonder if we should prefer methods over variables, as that retain more of the jnius current behaviour, by permitting include_protected and include_private to default to True.
> 
> I would be in favor of this. It's also an easy change to make.

Interesting, maybe it would be nice to have a way to specifically ask for an attribute over a method of the same name, but it's true it's a bit of a corner case, as it's certainly only possible if the attribute is private and the method is not? I not really sure of the best course, as i don't really have a java culture, i don't know what's common practice or not.

cmacdonald:
https://github.com/kivy/pyjnius/blob/master/jnius/jnius_export_class.pxi#L460 checks for obj is None. It should have checked for self.is_static.

Also writing of static fields was not implemented. I have a patch that addresses that, but it doesn't work  in the case of writing a field via its Child. I'm going to submit these as a separate PR, as they involve a different part of the code, and can be progressed separately.

cmacdonald:
> Interesting, maybe it would be nice to have a way to specifically ask for an attribute over a method of the same name, but it's true it's a bit of a corner case, as it's certainly only possible if the attribute is private and the method is not? 

I think we can be creative with names exposed in the python object:
- if a method exists with same name as a field (allowed in java), prefer the method and append _ to the field name
- if an attribute in child overrides parent, append super_ to the parent attribute. This is semantically equivalent to super.x in java
- it’s unlikely but if you have field (say x) with same name at different hierarchy levels, you might end up with super_super_x?

(But really, I just want a new release...)

hx2A:
> I think we can be creative with names exposed in the python object
> if a method exists with same name as a field (allowed in java), prefer the method and append _ to the field name

How do you do that? How do you make a field or method name different from what it is in the actual Java class? I thought they had to be the same for the reflection to work correctly and for it to find the right method/field.

This would be very useful for my situation, where I am working with a class with 6-7 fields & methods with the same name.

> https://github.com/kivy/pyjnius/blob/master/jnius/jnius_export_class.pxi#L460 checks for obj is None. It should have checked for self.is_static.

Is this fix checked in somewhere? Do you only have to change this one line of code to make it work? When I tried editing that line of code I still got seg faults.

cmacdonald:
> Is this fix checked in somewhere? Do you only have to change this one line of code to make it work? When I tried editing that line of code I still got seg faults.

See https://github.com/kivy/pyjnius/pull/505
You need both to compile and install
```
python setup.py bdist_wheel
pip install .
```

> How do you do that? How do you make a field or method name different from what it is in the actual Java class? I thought they had to be the same for the reflection to work correctly and for it to find the right method/field. This would be very useful for my situation, where I am working with a class with 6-7 fields & methods with the same name.

It is theoretically possible to separate the python and java names. But it does turn out to be fairly difficult. My work in progress can be seen at https://github.com/cmacdonald/pyjnius/commit/770863dff9e3a404269271f596496086cc006aeb. 
See also my comments on https://github.com/kivy/pyjnius/pull/505#issuecomment-611982191 for why it is difficult to implement.

hx2A:
I just pushed some code to make the `MetaJavaClass` class cache take into account the `include_protected` and `include_private` settings. I had to make some changes to `jnius_export_class.pxi`. Both Python metaprogramming and Cython are outside my core skill area so please give them a close look and provide feedback.

> Viz size, I wonder if we should prefer methods over variables, as that retain more of the jnius current behaviour, by permitting include_protected and include_private to default to True.

Currently, the two include_* params default to False and the code prefers variables over methods. Should I switch them to True and prefer methods over variables? It's an easy change to make and the unit tests still pass.

Once you two approve everything above, I'll finish up the unit tests this weekend. I want to test `include_protected` and `include_private` separately and I need to test how this works with inheritance.

> See #505

Great, thank you! I will give this a try.

> It is theoretically possible to separate the python and java names. But it does turn out to be fairly difficult.

This is probably a non-kosher hack, but I see I can make changes to non-static methods like so:

```
MyClass = autoclass('org.code.MyClass', include_protected=False, include_private=False)
MyClass.fake_name = MyClass.real_name  # new name for existing method
MyClass.special_protected_function = JavaMethod('()V')  # add protected method that autoclass left out
del MyClass.unwanted_function
myinstance = MyClass()

# myinstance is now a modified MyClass instance
myinstance.fake_name()
myinstance.special_protected_function()
assert not hasattr(myinstance, 'unwanted_function)
```

hx2A:
I added a lot more unit tests to properly test these features. I'm testing `include_protected` and `include_private` separately and am testing how this works with inheritance. I even added a test for `JavaMultipleMethods` with different visibility settings.

I believe this is just about done. I need @tshirtman and @cmacdonald to give a ruling on the default values for `include_protected` and `include_private` and preferring fields vs methods. Currently, the defaults are False and prefer fields over methods. Also, you should have a look at the cython/metaprogramming changes to verify I didn't break anything I don't understand.

hx2A:
Now that #505 is done, let's finish this one. Has all the rebasing and merging taken place? I'm happy to merge this into master.

Currently, the default values for `include_protected` and `include_private` are both False and the code prefers fields vs methods if a field and method have the same name.

Over the past week I made a lot of progress on my project that uses pyjnius, and I now believe that preferring methods over fields is the better choice. Like the `size()` method in `ArrayList`, classes can sometimes have private variables named identically to getter/setter methods they relate to. It is more important for those methods to be exposed, especially if those methods have some coding logic in them.

I prefer leaving the default values for `include_protected` and `include_private` as False.

@tshirtman @cmacdonald what do you think?

tshirtman:
My main gripe with that is it would possibly break existing programs. So i'd rather avoid it. I do think the temporary None + warning solution is acceptable.
Of course, size replacing size() is a bug we should fix. But maybe that can wait next release?
We could do a 1.3 now, for the new features, and a 2.0 soon after for this change?
Or would it be fine to do a 2.0 now, existing programs are going to have their code updated to use it anyway?

cmacdonald:
I would try to keep existing behaviour (by including protected and private) for a 1.3 release, but fix the size bug.

hx2A:
Existing behavior includes protected and private fields and methods in a class but excludes anything inherited, whether it be public or not or a field or method.

How about defaulting `include_protected` and `include_private` to `True` and preferring methods over fields (which would result in conflicting fields being hidden)?

tshirtman:
I think it would be good, hopefuly not too much code depends on the bug that makes private fields override the method name, but at worse they should be able to pin to the current version until they update their code.

hx2A:
Hashmap is another class that would have a problem with a field and method with the same name.

I just pushed a version that makes `include_protected` and `include_private` both default to `true` and prefers methods over fields.

Are we ready to merge?

cmacdonald:
Hi @hx2A ,

Do you check that HashMap, ArrayList have size as a method rather than a field - this would future-proof any future change breaking that. Something like:

```python
def assert_is_method(self, obj, name):
  multiple = "JavaMultipleMethod" in str(type(hm.__dict__['size']))
  single = "JavaMethod" in str(type(hm.__dict__['size']))
  self.assertTrue(multiple or single)

hm = autoclass("java.util.HashMap")
self.assert_is_method(hm, 'size')

al = autoclass("java.util.ArrayList")
self.assert_is_method(al, 'size')
```

That's all from me

hx2A:
@cmacdonald that's a good idea. I added a unit test that checks this and added a comment explaining what the unit test's purpose is.

cmacdonald:
thanks @hx2A . No further comments from me.

jnius/reflect.py
```diff
@@ -256,10 +260,37 @@ def autoclass(clsname):
...
+        for field in fields:
+            field_name = field.getName()
+            if field_name in cls_fields:
+                if level < cls_fields[field_name][1]:
+                    cls_fields[field_name] = (field, level)
+            else:
+                cls_fields[field_name] = (field, level)
+
+    # the fields are analyzed before methods so that if a method and a field
+    # have the same name, the field will take precidence in classDict.
```
tshirtman, line 281:
```suggestion
    # have the same name, the field will take precedence in classDict.
```

hx2A:
> Really nice, good job on the comments/docstrings/tests in particular, and the code is very clean.

Thank you, it means a lot.

> just a typo in a comment, (and you fixed an other typo so it evens out :))

haha!!

@cmacdonald @tshirtman it was a pleasure working with you both on this pull request. I am very happy to have made a contribution to this library.

cmacdonald:
thanks.

When doing something else, I found this:

```python
executor = autoclass("java.util.concurrent.Executors").newFixedThreadPool(1)
print(executor.__class__.__dict__['shutdown'])
```
returns
```
<property object at 0x104a63bd8>
```

and `executor.shutdown()` says `TypeError: 'bool' object is not callable`.

Could you give a look @hx2A?

Craig

hx2A:
Sure thing @cmacdonald 

Both the `shutdown` and `terminate` properties are added with this code:

```
if name != 'getClass' and bean_getter(name) and len(method.getParameterTypes()) == 0:
    lowername = lower_name(name[2 if name.startswith('is') else 3:])
    classDict[lowername] = (lambda n: property(lambda self: getattr(self, n)()))(name)
```

Why does `autoclass` add extra things to the class like this?

The [ExecutorService](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/ExecutorService.html) says there are methods `isShutdown()`, `isTerminated()`, `shutdown()`, and `terminated()`. The two methods that start with "is" are becoming properties and replace the other two methods.

These properties are put in `classDict` in the same loop that processes the `cls_methods` so I can't rearrange stuff the same way I did for fields. I believe these properties should have a lower priority than the methods (agreed?) and I can make this happen by putting them in second dict called `propertyDict` and then combining it with `classDict` before the protocol map loop.

Does that look like a good approach? I can make a new PR for this and add a few unit tests.

cmacdonald:
https://github.com/kivy/pyjnius/pull/464 seems to be related (@tshirtman). I'm not Python-esque enough to know why we need properties as well as JavaField. Cant we just check classDict first?

New PR would be good I think. Thanks!

tshirtman:
I'm not particularly convinced by this behavior either, i don't know how necessary it is, but i agree you could add these to a temporary dict, and merge them to classDict after the loop (unless there is collision, maybe dict.setdefaults would be nice to do that).

hx2A:
It has something to do with JavaBeans. I'm more into Python than Java, but I know some Java developers get excited about them.

hx2A:
See PR #518

PR #501, opened 2020-04-07
issue 465 - revised reflect

cmacdonald:
This addresses the recent discussion on https://github.com/kivy/pyjnius/issues/465#issuecomment-608997194

It also passes all pyjnius unit tests that I have currently.

### DIFF
--- a/jnius/reflect.py
+++ b/jnius/reflect.py
@@ -205,7 +205,7 @@ def identify_hierarchy(cls, level, concrete=True):
     for interface in interfaces or []:
         for sup, lvl in identify_hierarchy(interface, level + 1, concrete=concrete):
             yield sup, lvl
-    #all object extends Object, so if this top interface in a hierarchy, yield Object
+    # all object extends Object, so if this top interface in a hierarchy, yield Object
     if not concrete and cls.isInterface() and not interfaces:
         yield find_javaclass('java.lang.Object'), level +1
     yield cls, level
@@ -239,28 +239,28 @@ def autoclass(clsname):
 
     cls_methods=defaultdict(list)
 
-    #we now walk the hierarchy, from top of the tree, identifying methods
-    #hopefully we start at java.lang.Object 
+    # we now walk the hierarchy, from top of the tree, identifying methods
+    # hopefully we start at java.lang.Object 
     for cls,level in class_hierachy:
-        #dont analyse a given class more than once.
-        #many interfaces can lead to java.lang.Object 
+        # dont analyse a given class more than once.
+        # many interfaces can lead to java.lang.Object 
         if cls in cls_done:
             continue
         cls_done.add(cls)
-        #as we are walking the entire hierarchy, we only need getDeclaredMethods()
-        #to get what is in this class; other parts of the hierarchy will be found
-        #in those respective classes.
+        # as we are walking the entire hierarchy, we only need getDeclaredMethods()
+        # to get what is in this class; other parts of the hierarchy will be found
+        # in those respective classes.
         methods = cls.getDeclaredMethods()
         methods_name = [x.getName() for x in methods]
-        #collect all methods declared by this class of the hierarchy for later traversal
+        # collect all methods declared by this class of the hierarchy for later traversal
         for index, method in enumerate(methods):
             name = methods_name[index]
             cls_methods[name].append((cls, method, level))
     
-    #having collated the mthods, identify if there are any with the same name
+    # having collated the mthods, identify if there are any with the same name
     for name in cls_methods:
         if len(cls_methods[name]) == 1:
-            #uniquely named method
+            # uniquely named method
             owningCls, method, level = cls_methods[name][0]
             static = Modifier.isStatic(method.getModifiers())
             varargs = method.isVarArgs()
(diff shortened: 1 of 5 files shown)


jnius/reflect.py
```diff
@@ -217,58 +232,82 @@ def autoclass(clsname):
...
+        else:
             # multiple signatures
             signatures = []
-            for index, subname in enumerate(methods_name):
-                if subname != name:
+            #print("method with %d multiple signatures is %s in cls %s" % (len(clsMethods[name]), name, c))
+            
+            
+            #assume there can be no more than 10000 levels of inheritance
+            paramsig_to_level=defaultdict(lambda: 10000)
```
tshirtman, line 286: I don't think you need to assume that, you can use `float('inf')` as a starting point :).

jnius/reflect.py
```diff
@@ -195,6 +196,20 @@ def log_method(method, name, signature):
...
+    supercls = cls.getSuperclass()
+    if supercls is not None:
+        identifyHierarchy(supercls, level+1, classList, concrete)
+    intfcs = cls.getInterfaces()
+    if intfcs is not None:
+        for intf in intfcs:
+            identifyHierarchy(intf, level+1, classList, concrete)
+    if not concrete and cls.isInterface() and len(intfcs) ==0:
+        classList.append((find_javaclass('java.lang.Object'),level+1))
+    #return classList
```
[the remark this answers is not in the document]
cmacdonald: thanks for the timely feedback. I can make the changes. 

The yield looks more succinct - I dont know if it will work in Python 2 or not, so to be clear, do you wish this change?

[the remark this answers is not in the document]
tshirtman: i didn't test it, but i'm confident python2 shouldn't make a diff here, if we dropped python2 we could make it even more succint by replacing the inner loops with `yield from identify_hierarchy(_)`

cmacdonald:
Addressed feedback. There are a few extraneous commented #print() - do you want these left, removed or converted to debug?

tshirtman:
For the extra prints, well, either/or, I hadn't given a lot of though about logging them or not, if they are useful for debugging, it's better to leave them, i don't think the performance cost is that important here.

jnius/reflect.py
```diff
@@ -232,39 +232,36 @@ def autoclass(clsname):
...
 
-    classHierachy=[]
-    identifyHierarchy(c, 0, classHierachy, not c.isInterface())
-    classHierachy.reverse()
-
-    print("autoclass(%s) intf %r hierarchy is %s" % (clsname,c.isInterface(),classHierachy))
-    clsDone=set()
+    class_hierachy = identify_hierarchy(c, 0, not c.isInterface())
+    
+    log.debug("autoclass(%s) intf %r hierarchy is %s" % (clsname,c.isInterface(),class_hierachy))
```
tshirtman, line 237: that one might be less useful now it's a generator though :sweat_smile: maybe cast class_hierarchy to a list first to be able to both log and then iterate on it in the loop.

cmacdonald:
and finally, I added test cases for identify_hierarchy. I think this is as completed as I can.

tests/test_inheritance.py
```diff
@@ -1,10 +1,34 @@
...
     child = Child.newInstance()
+    parent = Parent.newInstance()
+    assert parent.PARENT_FIELD == 0
+    assert child.CHILD_FIELD == 1
+    assert child.PARENT_FIELD == 0
+    
 
+def test_newinstance():
+    import logging
+    logging.basicConfig(level=logging.DEBUG)
```
tshirtman, line 25: did you mean to leave that in?

tests/test_reflect.py
```diff
@@ -3,9 +3,43 @@
...
+        self.assertContains(d, "java.util.ArrayList")#concrete
+        self.assertContains(d, "java.util.AbstractCollection")#superclass
+        self.assertContains(d, "java.util.Collection")#interface
+        self.assertContains(d, "java.lang.Iterable")#interface
+        self.assertContains(d, "java.lang.Object")#root
+        maxLevel = max(d.values())
+        self.assertEqual(d["java.lang.Object"], maxLevel)
+        self.assertEqual(d["java.util.ArrayList"], 0)
+        
+
```
tshirtman, line 42: pep8, only one empty line between methods.

jnius/reflect.py
```diff
@@ -195,6 +196,20 @@ def log_method(method, name, signature):
...
+def identify_hierarchy(cls, level, concrete=True):
+    supercls = cls.getSuperclass()
+    if supercls is not None:
+         for sup, lvl in identify_hierarchy(supercls, level + 1, concrete=concrete):
+             yield sup, lvl # we could use yield from when we drop python2
+    interfaces = cls.getInterfaces()
+    for interface in interfaces or []:
+        for sup, lvl in identify_hierarchy(interface, level + 1, concrete=concrete):
+            yield sup, lvl
+    #all object extends Object, so if this top interface in a hierarchy, yield Object
```
tshirtman, line 208: pep8 nit, comments should start with `# ` not just `#`.

cmacdonald:
Thanks for the pep8 hints; c'est fait.

imagejan:
@tshirtman @cmacdonald Thanks for your work on `pyjnius`!
Is there a new release of `pyjnius` planned any time soon, including these changes?

cmacdonald:
Definitely, we're trying to get #500 landed - its very close.

PR #503, opened 2020-04-09
macOS - use /usr/libexec/java_home to detect Java

cmacdonald:
Avoids setting JAVA_HOME if possible on macOs

### DIFF
--- a/jnius/jnius_jvm_dlopen.pxi
+++ b/jnius/jnius_jvm_dlopen.pxi
@@ -2,7 +2,7 @@ include "config.pxi"
 import os
 from shlex import split
 from subprocess import check_output
-from os.path import dirname, join
+from os.path import dirname, join, exists
 from os import readlink
 from sys import platform
 from .env import get_jnius_lib_location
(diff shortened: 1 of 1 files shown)


jnius/jnius_jvm_dlopen.pxi
```diff
@@ -56,6 +56,19 @@ cdef find_java_home():
             except OSError:
                 break
         return dirname(dirname(java)).decode('utf8')
+    
+    if platform in ('darwin'):
```
tshirtman, line 60: The parens do nothing here, this would match 'win' as well as 'darwin'. either put `(darwin', )` or just use `== 'darwin'`.

jnius/jnius_jvm_dlopen.pxi
```diff
@@ -56,6 +56,19 @@ cdef find_java_home():
...
+        # its a mac
+        if not exists('/usr/libexec/java_home'):
```
tshirtman, line 62: Define this path as a local variable to avoid repetition.

cmacdonald:
feedback addressed

PR #505, opened 2020-04-09
reading static fields from instances, and writing to static fields

cmacdonald:
this addresses the reading of static fields from instances discussed in #500. 
It also addresses the writing to static fields, which was not implemented. 

Writing works when accessing the member variable itself, but the commented out lines in the test case do not, in that the the initial value of the field in child and Child are unchanged.
```python
    Parent.STATIC_PARENT_FIELD = 5
    assert Parent.STATIC_PARENT_FIELD == 5
    assert parent.STATIC_PARENT_FIELD == 5
    #assert Child.STATIC_PARENT_FIELD == 5
    #assert child.STATIC_PARENT_FIELD == 5
```

I dont think this is related to #500. 
Similarly, we can set in the child, and it doesn't propagate to the parent class.
```python
    Child.STATIC_PARENT_FIELD = 10
    assert Child.STATIC_PARENT_FIELD == 10
    assert child.STATIC_PARENT_FIELD == 10
    #assert Parent.STATIC_PARENT_FIELD == 10
    #assert parent.STATIC_PARENT_FIELD == 10
```

This clearly contrasts with Java, as per jshell:
```
$ CLASSPATH=build/classes/:build/test-classes jshell                          

jshell> org.jnius.Parent.STATIC_PARENT_FIELD = 5
$1 ==> 5

jshell> org.jnius.Child.STATIC_PARENT_FIELD
$2 ==> 5

jshell> org.jnius.Child.STATIC_PARENT_FIELD = 10
$3 ==> 10

jshell> org.jnius.Child.STATIC_PARENT_FIELD
$4 ==> 10
```

I'm wondering if _we_ need to keep track of the object that the static field belongs to...

So in short, this solves two bugs, but demonstrates a third. 
@tshirtman  @hx2A

### DIFF
--- a/jnius/jnius_export_class.pxi
+++ b/jnius/jnius_export_class.pxi
@@ -457,7 +457,7 @@ cdef class JavaField(object):
         cdef jobject j_self
 
         self.ensure_field()
-        if obj is None:
+        if self.is_static:
             return self.read_static_field()
 
         j_self = (<JavaClass?>obj).j_self.obj
@@ -467,9 +467,8 @@ cdef class JavaField(object):
         cdef jobject j_self
 
         self.ensure_field()
-        if obj is None:
-            # set not implemented for static fields
-            raise NotImplementedError()
+        if self.is_static:
+            self.write_static_field(value)
 
         j_self = (<JavaClass?>obj).j_self.obj
         self.write_field(j_self, value)
@@ -601,6 +600,57 @@ cdef class JavaField(object):
 
         check_exception(j_env)
         return ret
+
+    cdef write_static_field(self, value):
+        cdef jboolean j_boolean
+        cdef jbyte j_byte
+        cdef jchar j_char
+        cdef jshort j_short
+        cdef jint j_int
+        cdef jlong j_long
+        cdef jfloat j_float
+        cdef jdouble j_double
+        cdef jobject j_object
+        cdef JNIEnv *j_env = get_jnienv()
+
+        # type of the java field
+        r = self.definition[0]
+
+        # set the java field; implemented only for primitive types
+        if r == 'Z':
+            j_boolean = <jboolean>value
+            j_env[0].SetStaticBooleanField(j_env, self.j_cls, self.j_field, j_boolean)
+        elif r == 'B':
+            j_byte = <jbyte>value
+            j_env[0].SetStaticByteField(j_env, self.j_cls, self.j_field, j_byte)
+        elif r == 'C':
+            j_char = <jchar>value
+            j_env[0].SetStaticCharField(j_env, self.j_cls, self.j_field, j_char)
+        elif r == 'S':
+            j_short = <jshort>value
+            j_env[0].SetStaticShortField(j_env, self.j_cls, self.j_field, j_short)
+        elif r == 'I':
+            j_int = <jint>value
+            j_env[0].SetStaticIntField(j_env, self.j_cls, self.j_field, j_int)
+        elif r == 'J':
+            j_long = <jlong>value
+            j_env[0].SetStaticLongField(j_env, self.j_cls, self.j_field, j_long)
+        elif r == 'F':
+            j_float = <jfloat>value
+            j_env[0].SetStaticFloatField(j_env, self.j_cls, self.j_field, j_float)
+        elif r == 'D':
+            j_double = <jdouble>value
+            j_env[0].SetStaticDoubleField(j_env, self.j_cls, self.j_field, j_double)
+        elif r == 'L':
+            j_object = <jobject>convert_python_to_jobject(j_env, self.definition, value)
+            j_env[0].SetStaticObjectField(j_env, self.j_cls, self.j_field, j_object)
+            j_env[0].DeleteLocalRef(j_env, j_object)
+        else:
+            raise Exception(
+                "Invalid field definition '{}'".format(r)
+            )
+
+        check_exception(j_env)
 
     cdef read_static_field(self):
         cdef jboolean j_boolean
--- a/tests/java-src/org/jnius/Parent.java
+++ b/tests/java-src/org/jnius/Parent.java
@@ -2,6 +2,8 @@ package org.jnius;
 
 public class Parent {
 
+	public static int STATIC_PARENT_FIELD = 1;
+
 	public int doCall(Object o) {
 		return 0;
 	}
--- a/tests/test_inheritance.py
+++ b/tests/test_inheritance.py
@@ -1,7 +1,7 @@
-from jnius import autoclass
 
 
 def test_methodcalls():
+    from jnius import autoclass
     Parent = autoclass('org.jnius.Parent')
     Child = autoclass('org.jnius.Child')
     child = Child.newInstance()
@@ -11,6 +11,7 @@ def test_methodcalls():
     assert child.doCall(child) == 0
 
 def test_fields():
+    from jnius import autoclass
     Parent = autoclass('org.jnius.Parent')
     Child = autoclass('org.jnius.Child')
     child = Child.newInstance()
@@ -18,9 +19,27 @@ def test_fields():
     assert parent.PARENT_FIELD == 0
     assert child.CHILD_FIELD == 1
     assert child.PARENT_FIELD == 0
-    
+
+def test_staticfields():
+    from jnius import autoclass
+    Parent = autoclass('org.jnius.Parent')
+    Child = autoclass('org.jnius.Child')
+    child = Child.newInstance()
+    parent = Parent.newInstance()
+    assert Parent.STATIC_PARENT_FIELD == 1
+    assert Child.STATIC_PARENT_FIELD == 1
+    assert parent.STATIC_PARENT_FIELD == 1
+    assert child.STATIC_PARENT_FIELD == 1
+    #now test setting
+    Parent.STATIC_PARENT_FIELD = 5
+    assert Parent.STATIC_PARENT_FIELD == 5
+    assert parent.STATIC_PARENT_FIELD == 5
+    #assert Child.STATIC_PARENT_FIELD == 5
+    #assert child.STATIC_PARENT_FIELD == 5
+
 
 def test_newinstance():
+    from jnius import autoclass
     Parent = autoclass('org.jnius.Parent')
     Child = autoclass('org.jnius.Child')
 
@@ -29,4 +48,7 @@ def test_newinstance():
     assert isinstance(child, Parent)
 
 if __name__ == "__main__":
+    import jnius_config
+    jnius_config.add_options('-Xcheck:jni')
     test_newinstance()
+


tshirtman:
:fire: 

From a quick glance this seems correct, at least i don't see anything obviously wrong with it, but i must say i'm not really familiar with the usage of static in java, but from https://stackoverflow.com/questions/797964/what-is-the-exact-meaning-of-static-fields-in-java it seems we shouldn't have to do any tracking, as these two should always be the same object whatever we do? But maybe when the field is read/set, we need to lookup the inheritance chain to find the class it was declared in, and read/set it there? I'm not sure why it wouldn't crash when trying to set on a subclass, as it seems there shouldn't be any memory zone to write to here? Anyway, my understanding is probably wrong here, i'd need to read deeper into it to understand properly.

neither https://docs.oracle.com/javase/tutorial/java/IandI/subclasses.html or https://docs.oracle.com/javase/tutorial/java/javaOO/classvars.html seem to define how inheritance of static fields is supposed to work… there is a hint about overriding static methods in subclass to hide the original one, so i would assume static fields are inherited, and that they are shared accross all instances, whatever the subclass of the class that declares it, which would be consistent with your test with jshell, but i hoped to see it in writting somewhere (but i didn't read the whole pages, i searched for static and other keywords, maybe i missed something).

cmacdonald:
I have looked a bit more at this in the context of #500. Even to access a non-static field of the same name in a superclass, we would need to obtain the jclass id for all classes in the hierarchy, and then resolve fields with respect to the declaring object. 

As per https://github.com/kivy/pyjnius/blob/master/jnius/jnius_export_class.pxi#L207, we only keep the jclass for the actual class of the instance.

tshirtman:
Tried to bit, on top of your branch, i got the impression that we'd need to set the resolve info to the declaring class instead of the current class, so i did this

```diff
         # search all the static JavaField within our class, and resolve them
         cdef JavaField jf
+        cdef jclass declaring_cls
+        class_name = str_for_c(__javaclass__).replace('/', '.')
         for name, value in items_compat(classDict):
             if not isinstance(value, JavaField):
                 continue
             jf = value
             if not jf.is_static:
                 continue
-            jf.set_resolve_info(j_env, jcs.j_cls,
-                str_for_c(name), str_for_c(__javaclass__))
+            if jf.classname == class_name:
+                declaring_class = jcs.j_cls
+            else:
+                print("looking for declaring class %s of static field %s of %s" %
+                      (jf.classname, jf.name, class_name))
+                # declaring_cls = j_env[0].NewGlobalRef(
+                #     j_env,
+                j_env[0].FindClass(
+                    j_env,
+                    <char *>jf.classname
+                )
+                # )
+            jf.set_resolve_info(
+                j_env,
+                declaring_cls,
+                # jcs.j_cls,
+                str_for_c(name),
+                str_for_c(__javaclass__)
+            )
```
but it coredumps.
If i just do the lookup and don't use the result, i do get what seems to be correct information

```
looking for declaring class java.lang.Exception of static field serialVersionUID of java.lang.IndexOutOfBoundsException
looking for declaring class java.lang.RuntimeException of static field serialVersionUID of java.lang.IndexOutOfBoundsException
looking for declaring class java.lang.Throwable of static field $assertionsDisabled of java.lang.IndexOutOfBoundsException
```
Not sure where my understanding is incorrect, but it must be.

cmacdonald:
I tried to simplify, by quickly hacking up a mockup in C++ using the JNI interface. See attached file [FieldAccess.txt](https://github.com/kivy/pyjnius/files/4469242/FieldAccess.txt) (which can be renamed to .cc). On Mac, this compiles with 
```
g++ -fPIC -I"$JAVA_HOME/include" -I"$JAVA_HOME/include/darwin" -o jvmdemo FieldAccess.cc $JAVA_HOME/lib/server/libjvm.dylib
```
changing .dylib to .so and darwin to Linux (see https://www.developer.com/java/data/how-to-create-a-jvm-instance-in-jni.html). You would also need to update the classpath.

The output is
```
Parentfid is 4
  STATIC_PARENT_FIELD = 1
  STATIC_PARENT_FIELD = 200
Childfid is 4
  STATIC_PARENT_FIELD = 200
  STATIC_PARENT_FIELD = 200
```

This suggests that we're doing something wrong? I.e. we shouldn't need to track the declaring class.

cmacdonald:
Looking more at this. I dont think the problem is the field ids, etc. Its simply that JavaField.__set__() is not called when assigning a value to a static field. So once we have overwritten, Parent. STATIC_PARENT_FIELD is now a normal Python attribute. I haven't figured out why .__set__() is not called.

tshirtman:
that's not enough to fix the tests, but i did remark that the `__set__` method in your branch didn't return after setting the static field, so it also tried to set the non-static version of the field, but i'm indeed not sure this code is called, because it didn't crash (which i would expect) and the log i add there is not printed when the test fails.

tshirtman:
So, turns out that was a pretty good hint you got, took me some time to realize that setting the attribute on the class completely bypassed the descriptor. calling the `__set__` method could be achieved through the cumbersome, `Parent.__dict__['STATIC_PARENT_FIELD'].__set__(Parent, 5)` but of course that's not the api we want to implement.

So, how do we override the setting of a class attribute? Well, classes are objects too, and their class is a metaclass, so you implement `__setattr__` … on the metaclass. Here it's `MetaJavaClass`.

```
    def __setattr__(self, name, value):
        '''Overrides the setting *class* attributes, by overriding
        setattr on the metaclass (which classes are instance of), allows
        us to make the setting of static fields on the java class work.                                                                                                       
        '''
        log.debug('setting %s.%s = %s', self, name, value)
        if isinstance(self.__dict__.get(name), JavaStaticField):
            self.__dict__[name].__set__(self, value)
        else:
            super(MetaJavaClass, self).__setattr__(name, value)
```

I'll push the cleaned up commit in a branch if you want to pull it, as the other tests still fail.

tshirtman:
https://github.com/kivy/pyjnius/commit/a6713bf87334a13ae02935fdd674b8765a9ae072

cmacdonald:
Good job! I couldn't see what was wrong with `__set__()` last night. I can confirm that your work fixes the reading and writing to static fields. I have some enhanced test cases that use Java methods to ensure the values are written back to the Java object correctly.

On the other hand, between this and #500, we have three issues going on:
1. reading static fields from instances, and writing to static fields (#500)
2. protected/public and resolving clashing names properly using the inheritance hierarchy 
3. Rewriting clashing field names to allow the equivalent of super.field

(3) has proven to be difficult for me. Part of my confusion is that JavaField.name, .classname etc are sometimes bytes, and sometimes strings. 

Can this patch be committed independently of #500? If so, I think we should do so. Do you agree?

Effort can then resume to get #500 merged, and then create a separate issue for rewriting clashing fields IMHO, as the static fields are at least accessible via the owning Class, that could be prioritised lower.

tshirtman:
> Good job! I couldn't see what was wrong with `__set__()` last night. I can confirm that your work fixes the reading and writing to static fields. I have some enhanced test cases that use Java methods to ensure the values are written back to the Java object correctly.
> 
:+1: 

> (3) has proven to be difficult for me. Part of my confusion is that JavaField.name, .classname etc are sometimes bytes, and sometimes strings.

hm, that sounds like a bug, seems like we'd want to be as consistent as possible regarding that.

a propos of the super syntax for fields, since we are using descriptor to get and set already, so the java field never really changes, but is mutated instead, we could maybe have a `.super` method or attribute, instead of mangling names?

So instead of the java `object.super.name` we would have `object.name.super` if that makes sense. We would need to access the super class from the javafield, not sure how convenient that would be though, and i realize i don't know enough about how java implements that, (compared to python's `mro` logic).

> 
> Can this patch be committed independently of #500? If so, I think we should do so. Do you agree?
> 
> Effort can then resume to get #500 merged, and then create a separate issue for rewriting clashing fields IMHO, as the static fields are at least accessible via the owning Class, that could be prioritised lower.

I have nothing against that, this patch fixes issues, it doesn't depend on other stuff, we can certainly get it into the tree before. Do you want to do the rebase/cherry pick and force push to this branch or open a new PR? Otherwise i can do it later today.

cmacdonald:
Here's the revised test cases.

```python
def test_staticfields():
    from jnius import autoclass
    Parent = autoclass('org.jnius.Parent')
    Child = autoclass('org.jnius.Child')
    child = Child.newInstance()
    parent = Parent.newInstance()
    assert Parent.STATIC_PARENT_FIELD == 1
    assert Parent.StaticGetParentStaticField() == 1
    assert Child.STATIC_PARENT_FIELD == 1
    assert Child.StaticGetParentStaticField() == 1
    assert parent.STATIC_PARENT_FIELD == 1
    assert child.STATIC_PARENT_FIELD == 1
    assert parent.StaticGetParentStaticField() == 1
    assert child.StaticGetParentStaticField() == 1
    assert parent.getParentStaticField() == 1
    assert child.getParentStaticField() == 1

    # now test setting in parent
    Parent.STATIC_PARENT_FIELD = 5
    assert parent.STATIC_PARENT_FIELD == 5
    assert Child.STATIC_PARENT_FIELD == 5
    assert child.STATIC_PARENT_FIELD == 5
    assert Parent.StaticGetParentStaticField() == 5
    assert Child.StaticGetParentStaticField() == 5
    assert parent.StaticGetParentStaticField() == 5
    assert child.StaticGetParentStaticField() == 5
    assert parent.getParentStaticField() == 5
    assert child.getParentStaticField() == 5
    
    # now test setting in child
    child.STATIC_PARENT_FIELD = 10
    assert Child.STATIC_PARENT_FIELD == 10
    assert child.STATIC_PARENT_FIELD == 10
    assert Parent.STATIC_PARENT_FIELD == 10
    assert parent.STATIC_PARENT_FIELD == 10
    assert Parent.StaticGetParentStaticField() == 10
    assert Child.StaticGetParentStaticField() == 10
    assert parent.StaticGetParentStaticField() == 10
    assert child.StaticGetParentStaticField() == 10
    assert parent.getParentStaticField() == 10
    assert child.getParentStaticField() == 10
```

```java
package org.jnius;

public class Parent {

	public static int STATIC_PARENT_FIELD = 1;

	public int doCall(Object o) {
		return 0;
	}

	public int getParentStaticField() {
		return STATIC_PARENT_FIELD;
	}

	public static int StaticGetParentStaticField() {
		return STATIC_PARENT_FIELD;
	}

	public int getParentField() {
		return PARENT_FIELD;
	}

	static public Parent newInstance(){
		return new Parent();
	}

	public int PARENT_FIELD = 0;
	public int CLASH_FIELD = 0;
}
```

Here is the revised test case. It'll be this evening Europe time before I can rebase.

hx2A:
I'm happy to help merge #500 after you finish the rebase. I see the two PRs only have `jnius/jnius_export_class.pxi` in common and they change different parts of the file.

tshirtman:
I can do the rebase on my branch then :).

tshirtman:
i think this one can now be closed?

PR #506, opened 2020-04-11
implements more dunder protocols 

cmacdonald:
Inspired by https://github.com/kivy/pyjnius/issues/494#issuecomment-611511123, this PR has more support for Java Collections Framework, Iterators, Iterable and Comparable

I worked from https://rszalski.github.io/magicmethods/. Various test cases included.

Ones I didn't implement:
 - `__hash__(self)` - we couldnt guarentee the Python contract of hashCode() vs equals() wouldnt have been implemented on the Java side, as [Comparable](https://docs.oracle.com/javase/8/docs/api/java/lang/Comparable.html) doesnt require it.
 - `__nonzero__(self)` -   java.lang.Boolean (etc) are automapped  anyway
- `__str__(self)` - this could be mapped this to  toString(), but it was still handy in Python to know the jnius representation, i.e. that its a wrapper object.

I haven't verified on Python versions other than 3.6 and Java versions other than 13, so would appreciate the github workflow being run on this PR.

### DIFF
--- a/jnius/reflect.py
+++ b/jnius/reflect.py
@@ -339,17 +339,58 @@ def _getitem(self, index):
         else:
             raise
 
+#dunder method for java.util.Map
+def _map_getitem(self, k):
+    rtr = self.get(k)
+    if rtr is None:
+        raise KeyError()
+    return rtr
+
+#dunder method for java.util.Iterator
+def _iterator_next(self):
+    if not self.hasNext():
+        raise StopIteration()
+    return self.next()
+
+
 # protocol_map is a user-accessible API for patching class instances with additional methods 
 protocol_map = {
     'java.util.Collection' : {
-        '__len__' : lambda self: self.size()
+        '__len__' : lambda self: self.size(),
+        '__contains__' : lambda self, item: self.contains(item),
+        '__delitem__' : lambda self, item: self.remove(item)
     },
     'java.util.List' : {
         '__getitem__' : _getitem        
+    },
+    'java.util.Map' : {
+        '__setitem__' : lambda self, k, v : self.put(k,v),
+        '__getitem__' : _map_getitem,
+        '__delitem__' : lambda self, item: self.remove(item),
+        '__len__' : lambda self: self.size(),
+        '__contains__' : lambda self, item: self.containsKey(item),
+        '__iter__' : lambda self: self.keySet().iterator()
+    },
+    'java.util.Iterator' : {
+        '__iter__' : lambda self: self,
+        '__next__' : _iterator_next
+    },
+    'java.lang.Iterable' : {
+        '__iter__' : lambda self: self.iterator(),
     },
     # this also addresses java.io.Closeable
     'java.lang.AutoCloseable' : {
         '__enter__' : lambda self: self,
         '__exit__' : lambda self, type, value, traceback: self.close()
+    },
+    'java.lang.Comparable' : {
+        #__cmp__ is for Python 2 support
+        '__cmp__' : lambda self, other: self.compareTo(other),
+        '__eq__' : lambda self, other: self.equals(other),
+        '__ne__' : lambda self, other: not self.equals(other),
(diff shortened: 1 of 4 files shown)


jnius/reflect.py
```diff
@@ -339,17 +339,58 @@ def _getitem(self, index):
         else:
             raise
 
+#dunder method for java.util.Map
```
tshirtman, line 342: make these docstrings instead :)

tests/test_collections.py
```diff
@@ -0,0 +1,55 @@
+from __future__ import absolute_import
+import unittest
+from jnius import autoclass, protocol_map
+
+
+class TestCollections(unittest.TestCase):
+
+    def __init__(self, *args, **kwargs):
+        super(TestCollections, self).__init__(*args, **kwargs)
```
tshirtman, line 9: just curious why this pass-through override is needed? leftover from tests?

cmacdonald: yes. i developed the protocol_map in the test constructor before copying into reflect.py. do you want removed?

tshirtman: that would be appreciated, not a big deal, but good dead code is deleted code :P.

cmacdonald:
feedback address, thanks!

cmacdonald:
problems on Python 2 - see https://github.com/kivy/pyjnius/runs/579173024?check_suite_focus=true#step:9:138. Do you have a Python 2 dev environment @tshirtman ?

tshirtman:
woopsie, yes i do, i should have tested, thanks for the heads up!

tshirtman:
i can reproduce localy, let's try to make sense of it :]

cmacdonald:
https://github.com/kivy/pyjnius/runs/579173024?check_suite_focus=true#step:9:138
Python 2 needs `next()` rather than `__next__()`? (see https://stackoverflow.com/questions/41524531/python-2-vs-python-3-class-with-iter) The Java next() implementation wont raise StopIteration. 

Not sure how to fix this, as I'm not sure we can call "super().next()"?
  
https://github.com/kivy/pyjnius/runs/579173024?check_suite_focus=true#step:9:181 this line of code should be removed.

tshirtman:
yeah, it calls next() all the time, so the _iterator_next is not used, and so doesn't check for hasNext(). i think the last link is not what you wanted to paste.

cmacdonald:
Last link was not a fix for this, just an observation that a print(dir()) had made it into a test

We would need a special if() in autoclass for Python 2, which would need to rename the classDict entry for the Java next(). As we have discovered with the fields, renaming them on the Python side is tricky.

How long do you plan to support Python 2 for?

tshirtman:
> How long do you plan to support Python 2 for?

This is a strong reminder that we probably want to drop it next release :D (it would be a bit abrubt to drop without forewarning, right?) i'm myself would be perfectly fine dropping it asap, normaly everybody using pyjnius for android should have migrated to python3 anyway, as this one did drop support a couple releases ago.

> We would need a special if() in autoclass for Python 2, which would need to rename the classDict entry for the Java next(). As we have discovered with the fields, renaming them on the Python side is tricky.

Yeah, my ideas of how to fix this on the python side are quite gross, like

```
def _iterator_iter(self):
    print("starting iterator")
    if PY2:
        # in py2 the next() is called on the iterator, not __next__
        # so we need to monkey patch this one to check hasNext
        j_next = self.next
        log.debug("monkey patching next()")
        def safe_next():
            if not self.hasNext():
                raise StopIteration()
            return j_next()
        self.next = safe_next

    return self
```
(though it's not working yet in my tests, not sure why, it seems to recurse, and if it did, would that be fine?)

cmacdonald:
I can live with that implementation until Python 2 expires.

tshirtman:
so this works, and feels a bit less hacky https://github.com/kivy/pyjnius/pull/507

PR #507, opened 2020-04-11
Fix iterators py2

tshirtman:


tshirtman:
arf, of course i broke something in my cleanup :)

cmacdonald:
Thanks for sorting this out!

PR #508, opened 2020-04-12
Let the check_exception handler raise the error with more details.

kollivier:
I've been getting an error loading NativeInvocationHandler class using autoclass in the latest py3+webview bootstrap code for p4a, but the error message didn't contain many details. After investigating, I realized `check_exception` would give me those details, but the code was raising an exception manually before calling that function. 

This PR changes `find_javaclass` to return the exception generated by `catch_exception` instead of the currently thrown error.

### DIFF
--- a/jnius/jnius_export_func.pxi
+++ b/jnius/jnius_export_func.pxi
@@ -23,10 +23,6 @@ def find_javaclass(namestr):
     cdef JNIEnv *j_env = get_jnienv()
 
     jc = j_env[0].FindClass(j_env, name)
-    if jc == NULL:
-        j_env[0].ExceptionClear(j_env)
-        raise JavaException('Class not found {0!r}'.format(name))
-
     check_exception(j_env)
 
     cls = Class(noinstance=True)


cmacdonald:
Hi,
can you show the output with and without the patch?

kollivier:
With the patch:

04-11 22:08:54.480  3718  3718 I python  :   File "jnius/jnius_export_func.pxi", line 26, in jnius.jnius.find_javaclass
04-11 22:08:54.480  3718  3718 I python  :   File "jnius/jnius_utils.pxi", line 91, in jnius.jnius.check_exception
04-11 22:08:54.480  3718  3718 I python  : jnius.jnius.JavaException: JVM exception occurred: Didn't find class "org.jnius.NativeInvocationHandler" on path: DexPathList[[zip file "/data/app/org.learningequality.Kolibri-2/base.apk"],nativeLibraryDirectories=[/data/app/org.learningequality.Kolibri-2/lib/arm, /data/app/org.learningequality.Kolibri-2/base.apk!/lib/armeabi-v7a, /system/lib, /vendor/lib]]

Without the patch:

04-11 23:23:50.545 17653 17729 I python  :   File "jnius/jnius_export_func.pxi", line 28, in jnius.jnius.find_javaclass
04-11 23:23:50.545 17653 17729 I python  : jnius.jnius.JavaException: Class not found b'org/jnius/NativeInvocationHandler'

Notice that with the patch, it tells me the exact search path it used, which suggests that the issue is that the directory that contains `org.jnius.NativeInvocationHandler` is not on the search path, even though `jnius_config.get_classpath()` does include that dir.

BTW, since this problem appears to be specific to the webview bootstrap, I'm going to check into changes that were made to the SDL2 bootstrap that might not have been propagated or made in the same way to the webview bootstrap.

PR #518, opened 2020-04-22
ensure autoclass prefers methods over properties

hx2A:
Here is the PR for the problem with the ExecutorService Interface noted in #500.

Created one unit test to validate the JavaBean style properties do not replace a Java method.

### DIFF
--- a/jnius/reflect.py
+++ b/jnius/reflect.py
@@ -303,6 +303,8 @@ def autoclass(clsname, include_protected=True, include_private=True):
             if log.isEnabledFor(DEBUG):
                 log_method(method, name, sig)
             classDict[name] = (JavaStaticMethod if static else JavaMethod)(sig, varargs=varargs)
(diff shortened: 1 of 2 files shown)


cmacdonald:
Thanks, looks pretty good to me.
I wonder if we could take the opportunity to add a comment to explain what this logic is doing:
https://github.com/hx2A/pyjnius/blob/master/jnius/reflect.py#L306
making clear the linking of JavaBeans and Python properties.

hx2A:
>I wonder if we could take the opportunity to add a comment to explain what this logic is doing:
https://github.com/hx2A/pyjnius/blob/master/jnius/reflect.py#L306
making clear the linking of JavaBeans and Python properties.

It's always a good time for better comments.

I just added a comment explaining what that code is for.

cmacdonald:
thanks @hx2A, much appreciated!

Issue #428, opened 2019-06-27
libjvm.so is looked for in the wrong place


MaxNoe:
On travis, even with `JAVA_HOME` set, the library is not found.

It is actually in:

```
/usr/lib/jvm/java-8-openjdk-amd64/jre/lib/amd64/server/libjvm.so
```

But pyjnius looks in:

```
/usr/lib/jvm/java-8-openjdk-amd64/jre/lib/server/libjvm.so
```

julianhess:
You need to set `JDK_HOME` to `/usr/lib/jvm/java-8-openjdk-amd64/`; setup.py does not check `JAVA_HOME`. See:

https://github.com/kivy/pyjnius/blob/afd9b42ba54791870e789a18fbfed75bde97a609/setup.py#L154

MaxNoe:
Hey wo I need to set it at all? Why not take the java(c) that's in the PATH my default?

MaxNoe:
Oh wait. It does. Neither is the problem here!

The problem is the additional `amd64` in the path.

julianhess:
You need to rebuild the package with `JDK_HOME` already set as above; merely setting it after building will do nothing.

I kept running into this issue too until I dug through the source and found that the path to libjvm.so gets baked in during compilation (try running `strings jnius.cpython-37m-x86_64-linux-gnu.so | grep libjvm`)

tshirtman:
hopefully this should be solved by the changes i did in the #443 PR, which will provide wheels able to do this check at runtime, and try to be smarter about the lookup.

tshirtman:
Closing as i believe this to be fixed, let us know if you still have issues.

MaxNoe:
Just commenting: the PR you lined only changes a single line of whitespace. Wrong PR?

tshirtman:
indeed that was #444 thanks for pointing out.

mashabelyi:
I am having the same issue with the most recent version, even after setting both `JAVA_HOME` and `JDK_HOME` to my java path: `/usr/lib/jvm/java-8-oracle`.

In my case pyjnius is looking in 
```
/usr/lib/jvm/java-8-oracle/lib/server/libjvm.so
```

but the actual path is 
```
/usr/lib/jvm/java-8-oracle/jre/lib/amd64/server/libjvm.so
```

I tried adding `/usr/lib/jvm/java-8-oracle/jre/lib/amd64/server` to the `PATH` env variable too. But still get the same error on import:
```
[...]

PR #520, opened 2020-04-24
tentative fix for #428

tshirtman:


tshirtman:
no feedback, but this works here and should only be an improvement, merging.

PR #522, opened 2020-04-25
consider "package protected" methods and fields in autoclass

hx2A:
In addition to public, protected, and private visibility, methods and fields can be "package protected".

https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html

Package protected is the visibility setting when none of public, protected, and private are used (and therefore do not have any of those modifiers). The end result is the visibility depends on if the code is in the same package as the class package.

Rather than make a new `autoclass` parameter called `include_package_protected`, I made it look at the package names as it goes up the class hierarchy and follow `include_protected` if the package name is not the same as the starting package name. @tshirtman @cmacdonald I believe this is the better choice but am not 100% sure and want to hear what you have to say.

This PR also adds some unit test assert statements that should have been added in #500.

### DIFF
--- a/jnius/reflect.py
+++ b/jnius/reflect.py
@@ -95,12 +95,6 @@ class Modifier(with_metaclass(MetaJavaClass, JavaClass)):
     isTransient = JavaStaticMethod('(I)Z')
     isVolatile = JavaStaticMethod('(I)Z')
 
-    @classmethod
-    def isPackageProtected(cls, modifier):
-        return not (Modifier.isPublic(modifier) or
-                    Modifier.isProtected(modifier) or
-                    Modifier.isPrivate(modifier))
-
 
 class Method(with_metaclass(MetaJavaClass, JavaClass)):
     __javaclass__ = 'java/lang/reflect/Method'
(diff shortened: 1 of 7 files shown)


hx2A:
This might be better.

If a field or method is package protected:
  if `include_protected` is True and the package name is the same as the starting package name, include it. If `include_protected` is False, leave it out.
  if the package name is not the same, include it if `include_private` is True.

hx2A:
OK, now I am confident in what I have checked in.

For the input values `(include_protected, include_private)`, here is how package protected fields and method are handled:

* (True, True) => always include in class
* (True, False) => compare package of class in the hierarchy with the package `autoclass` started with, and only include if they are the same
* (False, False) => always exclude

Now all three possible outcomes are possible with different values for `(include_protected, include_private)`. None of the other proposed ideas can do that.

This should be considered a bug because right now the package protected fields and methods are always included, no matter what `(include_protected, include_private)` are set to.

jnius/reflect.py
```diff
@@ -95,6 +95,12 @@ class Modifier(with_metaclass(MetaJavaClass, JavaClass)):
     isTransient = JavaStaticMethod('(I)Z')
     isVolatile = JavaStaticMethod('(I)Z')
 
+    @classmethod
+    def isPackageProtected(cls, modifier):
```
tshirtman, line 99: i'm not sure about adding python code like this on a java class, i agree it's convenient, makes calls consistent with others, but it feels like a hack, as it's not a method of the java class, afaik. Will wait a bit to see if i come to term with the idea or not.

hx2A: When I added that I both thought it was clever and didn't like it because I'm making it different from what the actual Java `Modifier` class is like. This new method is only used in two places, so it doesn't actually add that much value anyway. I took it out.

tshirtman: yeah, it's not a big code duplication anyway, i think it's better like this, thanks.

PR #523, opened 2020-04-26
expose the class's class attribute

cmacdonald:
In Java, I can use `String.class` to get the String class' Class object. 
In Jnius, its not exposed. This fixes this.

Note that as `class` is a reserved keyword in Python, the attribute is `._class`, so the python equivalent is `autoclass("java.lang.String")._class`.

Includes a unit test.

### DIFF
--- a/tests/test_reflect.py
+++ b/tests/test_reflect.py
@@ -38,6 +38,15 @@ class ReflectTest(unittest.TestCase):
         self.assertEqual(d["java.lang.Object"], maxLevel)
         self.assertEqual(d["java.util.ArrayList"], 0)
 
+    def test_class(self):
+        lstClz = autoclass("java.util.List")
+        self.assertTrue("_class" in dir(lstClz))
+        self.assertEqual("java.util.List", lstClz._class.getName())
+        alstClz = autoclass("java.util.ArrayList")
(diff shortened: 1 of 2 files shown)


tshirtman:
At first i though "hey, but in python it's `__class__` so we could expose it as that, would be neat", but then uh, no, that's almost certainly a terrible idea :laughing:.
Having a look :)

cmacdonald:
`__javaclass__` has the string of the classname, but there was no way to get the Class object without instantiating the object (or using `find_javaclass()`)

tests/test_reflect.py
```diff
@@ -38,6 +38,15 @@ def test_hierharchy_arraylist(self):
         self.assertEqual(d["java.lang.Object"], maxLevel)
         self.assertEqual(d["java.util.ArrayList"], 0)
 
+    def test_class(self):
+        lstClz = autoclass("java.util.List")
```
tshirtman, line 42: that's a nothingbug but in python the standin for "class" is usually "cls" (and sometime "klass" but i'm less fond of that). but here it could even be the more idiomatic "list_class".

cmacdonald: fair. naming was not very pythonic.

Issue #437, opened 2019-08-11
ValueError: chr() arg not in range(0x110000)


Chriskamphuis:
`JString` does not seem to handle all characters correctly:

This code snippet will result in the error:
```
from jnius import autoclass
JString = autoclass('java.lang.String')
JString('ö')
```

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "jnius/jnius_export_class.pxi", line 256, in jnius.JavaClass.__init__
  File "jnius/jnius_export_class.pxi", line 341, in jnius.JavaClass.call_constructor
  File "jnius/jnius_conversion.pxi", line 24, in jnius.release_args
  File "jnius/jnius_conversion.pxi", line 288, in jnius.convert_jarray_to_python
ValueError: chr() arg not in range(0x110000)
```

Chriskamphuis:
This solves the problem:

`JString('ö'.encode('utf-8'))`

But I do not know if this behavior is desired?

hx2A:
I'm working on a fix for this in PR #524.

hx2A:
@tshirtman I think you can close this issue now because of #524 . One of my unittests verifies this specific thing.

tshirtman:
great, thanks :)

PR #524, opened 2020-04-30
remove <char> casts to allow for than 256 characters

hx2A:
Java `char` [datatypes](https://docs.oracle.com/javase/tutorial/java/nutsandbolts/datatypes.html) have a maximum value of `0xFFFF` but Python `chr` go up to `0x110000`. But in `jnius_export_class.pxi` there are `<char>` casts like `chr(<char>j_char)` to read char fields or  char method return values. But this is not correct if `j_char` is greater than 255. Taking out the casts allows for more characters.

### DIFF
--- a/jnius/jnius_conversion.pxi
+++ b/jnius/jnius_conversion.pxi
@@ -286,7 +286,7 @@ cdef convert_jarray_to_python(JNIEnv *j_env, definition, jobject j_object):
         j_chars = j_env[0].GetCharArrayElements(
                 j_env, j_object, &iscopy)
 
-        if sys.version_info < (3, 0):
+        if PY_MAJOR_VERSION < 3:
             ret = [chr(<char>j_chars[i]) for i in range(array_size)]
         else:
             ret = [chr(j_chars[i]) for i in range(array_size)]
--- a/jnius/jnius_conversion.pxi
+++ b/jnius/jnius_conversion.pxi
@@ -285,7 +285,11 @@ cdef convert_jarray_to_python(JNIEnv *j_env, definition, jobject j_object):
     elif r == 'C':
         j_chars = j_env[0].GetCharArrayElements(
                 j_env, j_object, &iscopy)
-        ret = [chr(<char>j_chars[i]) for i in range(array_size)]
+
+        if sys.version_info < (3, 0):
+            ret = [chr(<char>j_chars[i]) for i in range(array_size)]
+        else:
+            ret = [chr(j_chars[i]) for i in range(array_size)]
         j_env[0].ReleaseCharArrayElements(
                 j_env, j_object, j_chars, 0)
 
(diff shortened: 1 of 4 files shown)


hx2A:
Didn't actually test Python 2.7 yet though, let me do that now.

hx2A:
Since in Python 2.7 `char` has a maximum value of `0xFF`, this fix should only apply for Python 3. I added a check in the code to skip the `<char>` cast for Python 3. The unit tests only test higher byte characters and strings in Python 3.

hx2A:
Issue #437 tells me I have more work to do, and I now have a fix for it.

I will create more unit tests and hunt around in `jnius_conversion.pxi` for the <char> casts.

hx2A:
Here's the summary of what this is about: Java characters use UTF-16 and have a range of 0xFFFF. Python 2.7 characters had a range of 0xFF. The `<char>` cast also limits the range to 0xFF and needed to be there to prevent an error. Now that Python 3 has a range of 0x110000, the `<char>` cast is not necessary and throttles the range to 0xFF, limiting the characters.

For higher byte characters the `<char>` cast sometimes results in a negative number, which causes an error. Other times the higher byte character will be changed into a different character in the range 0xFF.

This fix removes the `<char>` cast for Python 3.

@tshirtman I don't see `<char>` used anywhere else in the code and think this is ready for your review. I believe this will also fix a bunch of open issues.

cmacdonald:
Hi Jim,

For detecting Python version, see https://github.com/kivy/pyjnius/search?q=PY_MAJOR_VERSION&unscoped_q=PY_MAJOR_VERSION

I think that will be faster than checking sys.version_info?

tshirtman:
I was a bit unsure about that, because if it's a char/str in python2, shouldn't it be a byte in python3? But if java treats it as utf16 encoded, then obviously it's not the case. I'm still not completly sure about the implications of this patch, but i guess the current situation is wrong anyway, and it looks like at least an improvement.

I don't think the platform check should make a big hit on performance, but sure, we can certainly use cython's method if so.

hx2A:
I changed the version check to `PY_MAJOR_VERSION < 3`, which also makes it consistent with what is done elsewhere in the cython code.

hx2A:
What's the "hx2A dismissed tshirtman’s stale review via 7c12252 1 minute ago" github message?

tshirtman:
That (confusing, imho) message from github just indicates that your changes make my review obsolete, so it's dismissed. It makes it sounds much more like a conscious decision on your side that it usually is though.

AndreMiras:
This PR breaks tests on Windows it seems :confused: 
https://github.com/kivy/pyjnius/runs/637247538?check_suite_focus=true
```
    [javac] D:\a\pyjnius\pyjnius\tests\java-src\org\jnius\CharsAndStrings.java:15: error: unclosed character literal
    [javac]   static public char testStaticChar2 = 'ä';
```
It's like Windows is taking it as byte and expect it to be of string type and double quoted, what do you think guys?

tshirtman:
source file encoding issue? https://stackoverflow.com/a/1726222/1209937 I guess our file is encoded in utf8 (as normal), but windows doesn't default to that, so it's not read properly by javac. Maybe adding the -encoding parameter when building the tests source would fix it.

Will push a PR to try that.

PR #535, opened 2020-05-09
micro typo fix for dlopen failures

slippycheeze:


slippycheeze:
Note: I'm highly unlikely to check back in on this pull request, just noticed the issue in passing and figured I'd help out trivially. :)

tshirtman:
Thanks for trying to help, but i don't think that's correct, the parenthesis you are trying to close is already closed at the end of the line, and you are closing it inside the string, which has no effect anyway.

PR #542, opened 2020-05-31
Improve performance of byte array parameters

hx2A:
Here is my pull request for my work so far to improve the `jnius.ByteArray` class with the end goal of making it faster to pass large blocks of raw bytes back and forth between Python and Java.

Raw bytes can be in three forms: the `bytes` type, i.e. `b'123'`, the builtin `bytearray` type, i.e. `bytearray(b'123')`, and the class `jnius.ByteArray`. The `ByteArray` class is created by the jnius library when Java functions have the `byte[]` return type and currently cannot be created adhoc (something like `jnius.jnius.ByteArray(b'123')` will execute but the ByteArray buffer will be empty).

I need to pass blocks of bytes back and forth between Python and Java. Through experimenting I discovered I could pass a `ByteArray` instance back to Java much faster than passing a `bytearray` instance with the same size. I also saw I could fiddle with the `ByteArray` code to make it mutable, providing me with the opportunity to quickly get a `ByteArray`, make the changes I needed, and then pass it back. This works well for my purposes.

The new `__setitem__` method I added to jnius_nativetypes.pxi seems to work but probably has some bugs. Obviously I'll create unittests for this. I'd like to try out different approaches for `__setitem__` and I might as well make improvements to `__getitem__` also while I'm at it.

On discord I was commenting to @tshirtman that merely adding the `__getitem__` method to `ByteArray` hurt performance, even if I wasn't actually using it in my code. This was due to a combination of two problems: first, my `__getitem__` method was slow, and second, the `release_args` function in jnius_conversion.pxi. In `release_args` there's a block of code that does this:

```
            try:
                args[index][:] = ret
            except TypeError:
                pass
```

Before I added the `__getitem__` method, when `ByteArray` objects hit this code the exception would be thrown and no copying would happen. After I added it, the exception would not be thrown and the `__getitem__` method which at the time was slow would be used to recopy the data, huring performance.

I understand what this is for: it is to simulate passing objects by reference. (I figured it out by commenting this out and seeing which unittests broke :) ) Since Python and Java don't share the same memory, objects must be passed by value. To simulate passing by reference, they are copied back after the function completes.

Simulating pass by reference makes sense for most programming situations because it is what programmers would expect. However, in my case of passing large blocks of bytes, it is an extra memory copy that I'd rather avoid. Therefore, I'd like to propose a way to make that second copy optional, either for `ByteArray`s only  or other Java arrays as well. In this PR I added a `_JNIUS_PASS_BY_VALUE` attribute to the `ByteArray` class that the `release_args` function checks for. Although I see that what I wrote doesn't work for anything other than `ByteArray` though, so this needs some work.

This PR is a work in progress but I wanted @tshirtman , @cmacdonald and anyone else to be able to comment on what I've done so far. Also it happens I have surgery tomorrow and will need to be away from my computer for a while, so these notes will be helpful to me when I pick this up again.

Things to do and questions:

* Learn Cython
* Performance test Cython code for the `__getitem__` and `__setitem__` methods to make them as fast as possible
* Should I be adding a `__setitem__` method at all? How about a `replace_data` method instead?
* Any data updating would need to have checks in place to make sure the new data has the right length and probably a bunch of other things
* Should I make similar changes to jnius_nativetypes.pxi (for Python 2.7)?
* Revisit the performance of passing the builtin type `bytearray`. Can it be made to be as fast as `jnius.ByteArray`? Why isn't it as fast?
* Unittests. Lots of unittests.

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -171,10 +171,14 @@ Reflection classes
 Reflection functions
 --------------------
 
-.. function:: autoclass(name)
+.. function:: autoclass(name, include_protected=True, include_private=True)
 
     Return a :class:`JavaClass` that represents the class passed from `name`.
     The name must be written in the format `a.b.c`, not `a/b/c`.
+
+    By default, autoclass will include all fields and methods at all levels of
+    the inheritance hierarchy. Use the `include_protected` and `include_private`
+    parameters to limit visibility.
 
     >>> from jnius import autoclass
     >>> autoclass('java.lang.System')
@@ -184,6 +188,10 @@ Reflection functions
 
     >>> autoclass('android.provider.Settings$Secure')
     <class 'jnius.reflect.android.provider.Settings$Secure'>
+
+    .. note::
+        If a field and a method have the same name, the field will take
+        precedence.
 
     .. note::
         There are sometimes cases when a Java class contains a member that is
@@ -223,7 +231,7 @@ Java class implementation in Python
     You need to define at minimum the :data:`__javainterfaces__` attribute, and
     declare java methods with the :func:`java_method` decorator.
 
-    .. notes::
+    .. note::
 
         Static methods and static fields are not supported.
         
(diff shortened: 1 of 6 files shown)


hx2A:
I've implemented the kwarg syntax and like this approach a lot better. It seems to work correctly for constructor parameters and method parameters.

If you like this I'll start on the unit tests and add your list idea, which I think is worthwhile:
```
obj.method(int_list, string, pass_by_reference=[True, False])
```

cmacdonald:
I agree it looks like a convincing method.

We need a block of documentation about these "magic flags". `noinstance` is jnius for internal use, but both `signature` and `pass_by_reference` are developer facing. When should a developer pass by reference or pass by value.

hx2A:
Sure, I'll write some documentation too. I see the documentation is in the docs/source directory. I'll do that with the unit tests.

hx2A:
I made it so the `pass_by_reference` kwarg can be a boolean or a list/tuple. If the list/tuple is shorter than the number of params, the last value will be used for the remainder of the parameters.

```
In [3]: numbers = [list(range(i)) for i in range(3, 7)]

In [4]: numbers
Out[4]: [[0, 1, 2], [0, 1, 2, 3], [0, 1, 2, 3, 4], [0, 1, 2, 3, 4, 5]]

In [5]: obj.testIntMutability(*numbers, pass_by_reference=[False, True, False])

In [6]: numbers
Out[6]: [[0, 1, 2], [13, 1, 2, 3], [0, 1, 2, 3, 4], [0, 1, 2, 3, 4, 5]]

In [7]: obj.testIntMutability(*numbers, pass_by_reference=True)

In [8]: numbers
Out[8]: [[13, 1, 2], [13, 1, 2, 3], [13, 1, 2, 3, 4], [13, 1, 2, 3, 4, 5]]
```

I know typing variables is important for Cython performance but I am not sure if I made the best choices. Let me know if I should change what I just checked in.

hx2A:
Unit tests for the `pass_by_reference` param are complete. Should I add unit tests for anything else? The speed of passing large byte parameters?

I'll do the documentation next.

hx2A:
Documentation is done. I also added docs for the previously made enhancements to `autoclass`. Anything else that needs to be done?

I just counted and I see I've made a total of 50 unit tests for Pyjnius. Ironically, @tshirtman 's initial feedback on my first PR was that I had no unit tests. :)

tshirtman:
Thank you for your fantastic work! And sorry I didn't give it more attention, things have been a bit busy in life lately, and will for some more time, but hopefuly it'll be better in a couple of weeks.

Onto reviewing!

Issue #562, opened 2020-10-22
Failed to find constructor when there are multiple(overloaded) constructors


enjoybeta:
Found a bug with pyjnius when dealing with a Java class with multiple constructors.

When I executed the below Java code, the main method will invoke the 2nd constructor and finish gracefully.
When I execute the python code below, it will report the constructor is not found.
`(jnius.JavaException: No constructor matching your arguments, available: ['(ILjava/lang/String;)V', '(ILjava/lang/String;Ljava/lang/Object;[I)V'])`
But when I removed the 1st constructor in java code, my python code with pyjnius can correctly find the only constructor and finish gracefully.

**Java Class**
```
public class SampleJavaClass {
    public SampleJavaClass( int arg1, String arg2, int arg3, Object arg4, int... arg5 ) {
	System.out.println("arg1: " + Integer.toString(arg1));
        System.out.println("arg2: " + arg2);
        System.out.println("arg3: " + arg3);
        System.out.println("arg3: " + arg4);
        System.out.println("arg4: " + arg5.toString());
    }

    // the constructor plans to use
    public SampleJavaClass (int arg1, String arg2, Object arg3, int... arg4) {
        System.out.println("arg1: " + Integer.toString(arg1));
        System.out.println("arg2: " + arg2);
        System.out.println("arg3: " + arg3);
        System.out.println("arg4: " + arg4.toString());
    }
    
    public static void main(String[] args) {
        SampleJavaClass test = new SampleJavaClass(1, "var2", null, 4);
    }
}
```

**Python code with pyjnius**
```
import os
currentPath = os.getcwd()
classpath = currentPath + "/SampleJavaClass.class"
import jnius_config
jnius_config.set_classpath('.', classpath)
from jnius import autoclass

def main():
    SampleJavaClass = autoclass("SampleJavaClass")
    SampleJavaClass(1, "var2", None, 4)
    print("Execution finished!")


if __name__ == "__main__":
    main()
```

cmacdonald:
have you tried using a constructor signature hint?

See https://github.com/kivy/pyjnius/blob/307659b13c1e5583fcb25603b7d3732265ffd4a0/tests/test_constructor.py#L54 for an example.

hx2A:
[...]

PR #566, opened 2020-10-22
fix bug for constuctors with variable arguments

hx2A:
This is a fix for #562 

Simple fix so that `calculate_score` works for constructors the way it works for methods.

### DIFF
--- a/jnius/jnius_export_class.pxi
+++ b/jnius/jnius_export_class.pxi
@@ -338,8 +338,10 @@ cdef class JavaClass(object):
                 )
             if not scores:
                 raise JavaException(
-                    'No constructor matching your arguments, available: '
-                    '{}'.format(found_definitions)
+                    'No constructor matching your arguments, requested: {}, available: {}'.format(
+                        args,
+                        found_definitions
+                    )
                 )
             scores.sort()
             score, definition, d_ret, d_args, args_ = scores[-1]
--- a/jnius/jnius_export_class.pxi
+++ b/jnius/jnius_export_class.pxi
@@ -327,7 +327,7 @@ cdef class JavaClass(object):
                 else:
                     args_ = args
 
-                score = calculate_score(d_args, args)
+                score = calculate_score(d_args, args_, is_varargs)
                 if score == -1:
                     continue
                 scores.append((score, definition, d_ret, d_args, args_))
--- a/jnius/jnius_export_class.pxi
+++ b/jnius/jnius_export_class.pxi
@@ -341,7 +341,8 @@ cdef class JavaClass(object):
                 )
             if not scores:
                 raise JavaException(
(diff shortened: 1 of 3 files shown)


cmacdonald:
Good spot Jim. I realise that while I recently improved the error messages for methods,
c.f. https://github.com/kivy/pyjnius/blob/master/jnius/jnius_export_class.pxi#L341
I didnt do likewise here:
https://github.com/kivy/pyjnius/blob/master/jnius/jnius_export_class.pxi#L1143

Could we fix that too?

tshirtman (approved):
Nice catch!

hx2A:
@tshirtman thanks!

I just added to the PR to improve the error message, as per @cmacdonald 's suggestion. Example error is below. Is it possible to further improve this to also say the class name (like the similar error message for methods)? Would that even be useful?

```
In [7]: String = jnius.autoclass('java.lang.String')

In [8]: String(1, 2, 3, 'asdfasdf')
---------------------------------------------------------------------------
JavaException                             Traceback (most recent call last)
<ipython-input-8-327bc0e8194d> in <module>
----> 1 String(1, 2, 3, 'asdfasdf')

~/Projects/ITP/pythonprocessing/pyjnius/jnius/jnius_export_class.pxi in jnius.JavaClass.__init__()
    268 
    269         if 'noinstance' not in kwargs:
--> 270             self.call_constructor(args, kwargs)
    271             self.resolve_methods()
    272             self.resolve_fields()

~/Projects/ITP/pythonprocessing/pyjnius/jnius/jnius_export_class.pxi in jnius.JavaClass.call_constructor()
    338                 )
    339             if not scores:
--> 340                 raise JavaException(
    341                     'No constructor matching your arguments, requested: {}, available: {}'.format(
    342                         args,

JavaException: No constructor matching your arguments, requested: (1, 2, 3, 'asdfasdf'), available: ['([B)V', '([BII)V', '([BLjava/nio/charset/Charset;)V', '([BLjava/lang/String;)V', '([BIILjava/nio/charset/Charset;)V', '(Ljava/lang/StringBuilder;)V', '(Ljava/lang/StringBuffer;)V', '([CII)V', '([C)V', '(Ljava/lang/String;)V', '()V', '([BIILjava/lang/String;)V', '([BI)V', '([BIII)V', '([III)V']
```

hx2A:
I just added an additional fix for signature hints on constructors with variable arguments. Previously there was an assert statement that forbid the use of signature hints on constructors with variable arguments.

tshirtman:
> Example error is below. Is it possible to further improve this to also say the class name (like the similar error message for methods)? Would that even be useful?

i think you can use `self.__javaclass__` ? and yes, that would certainly be an improvement.

hx2A:
@tshirtman that worked

tshirtman (approved):

Issue #538, opened 2020-05-22
1.3.0 breaks android's request_permission


tshirtman:
reported by an user (Matrix0206 on discord).

```
from android import request_permission, Permission

request_permissions([Permission.READ_EXTERNAL_STORAGE, Permission.WRITE_EXTERNAL_STORAGE])
```

```
05-22 20:08:20.301 10976 11090 I python  :  Traceback (most recent call last):
05-22 20:08:20.301 10976 11090 I python  :    File "/run/media/matrix/New_Volume/Developments/kivy-music-player/.buildozer/android/app/main.py", line 300, in <module>
05-22 20:08:20.301 10976 11090 I python  :    File "/run/media/matrix/New_Volume/Developments/kivy-music-player/.buildozer/android/platform/build-arm64-v8a/build/python-installs/music_player/kivy/app.py", line 949, in run
05-22 20:08:20.302 10976 11090 I python  :    File "/run/media/matrix/New_Volume/Developments/kivy-music-player/.buildozer/android/platform/build-arm64-v8a/build/python-installs/music_player/kivy/app.py", line 919, in _run_prepare
05-22 20:08:20.302 10976 11090 I python  :    File "/run/media/matrix/New_Volume/Developments/kivy-music-player/.buildozer/android/app/main.py", line 132, in build
05-22 20:08:20.302 10976 11090 I python  :    File "/run/media/matrix/New_Volume/Developments/kivy-music-player/.buildozer/android/platform/build-arm64-v8a/build/python-installs/music_player/kivy/uix/relativelayout.py", line 274, in __init__
05-22 20:08:20.303 10976 11090 I python  :    File "/run/media/matrix/New_Volume/Developments/kivy-music-player/.buildozer/android/platform/build-arm64-v8a/build/python-installs/music_player/kivy/uix/floatlayout.py", line 65, in __init__
05-22 20:08:20.303 10976 11090 I python  :    File "/run/media/matrix/New_Volume/Developments/kivy-music-player/.buildozer/android/platform/build-arm64-v8a/build/python-installs/music_player/kivy/uix/layout.py", line 76, in __init__
05-22 20:08:20.303 10976 11090 I python  :    File "/run/media/matrix/New_Volume/Developments/kivy-music-player/.buildozer/android/platform/build-arm64-v8a/build/python-installs/music_player/kivy/uix/widget.py", line 359, in __init__
[...]

PR #567, opened 2020-10-22
Fix static methods

obfusk:
Properly fixes the issue found via #538.

This was already partially fixed between 1.2.1 and 1.3.0 because the new class hierarchy walk resulted in more `JavaMultipleMethod`s, which already handled static methods correctly.

### DIFF
--- a/jnius/jnius_export_class.pxi
+++ b/jnius/jnius_export_class.pxi
@@ -321,7 +321,7 @@ cdef class JavaClass(object):
                     score=1
                     scores.append((score, definition, d_ret, d_args, args))
                     break
-                
+
                 if is_varargs:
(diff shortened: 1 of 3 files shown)


cmacdonald:
The patch itself looks good. I was thinking about:
`self.assertTrue(inst.getClass().isInstance(inst))`.
To retain also the original intention of the test, could you also add:
`self.assertTrue(ConstructorTest._class.isInstance(inst))`

obfusk:
Thanks! I was looking for `._class` to be able to do that :)

cmacdonald (approved):
I approve! Thanks!

tshirtman (approved):
nice :)

PR #568, opened 2020-10-22
set_resolve_info: pass j_self instead of None

### LINKED ISSUE #557: set_resolve_info passes None instead of j_self
In this code `j_self` is passed to `set_resolve_info` when it is `None` and `None` is passed when it is not.

https://github.com/kivy/pyjnius/blob/3371e1012873fe373773c43b556d2796d10d102e/jnius/jnius_export_class.pxi#L1097-L1109

Not sure if it's a bug because the control flow is a bit complex here.

obfusk:
Fixes #557.

### DIFF
--- a/jnius/jnius_export_class.pxi
+++ b/jnius/jnius_export_class.pxi
@@ -1098,14 +1098,14 @@ cdef class JavaMultipleMethod(object):
                 if signature in self.static_methods:
                     continue
                 jm = JavaStaticMethod(signature, varargs=is_varargs)
-                jm.set_resolve_info(j_env, j_cls, j_self, name, classname)
+                jm.set_resolve_info(j_env, j_cls, None, name, classname)
                 self.static_methods[signature] = jm
 
             elif j_self is not None and not static:
                 if signature in self.instance_methods:
                     continue
                 jm = JavaMethod(signature, varargs=is_varargs)
-                jm.set_resolve_info(j_env, j_cls, None, name, classname)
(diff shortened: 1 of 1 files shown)


cmacdonald:
But is there a (new) unit test that can exercise this change?

obfusk:
> But is there a (new) unit test that can exercise this change?

Since the bug is currently not actually triggerable unless you do really weird stuff (since `__get__` always sets `j_self`, in this case not redundantly), that seems like overkill.

cmacdonald:
So can we think of what actually triggers this? I cant reasonably validate its actually a bug or see why it needs to be fixed if we have no way of reproducing.

I do do weird thinks like adding dunder methods to Java objects.

obfusk:
It's a bug in the sense that it leaves an object in an inconsistent state. But because there is currently no (reasonable) code path that leads to the state still being inconsistent when it matters, it can't really be reproduced or triggered.

But if such a code path is later added (based on the assumption that the state is consistent)...

obfusk:
```py
>>> from jnius import autoclass
>>> s = autoclass('java.lang.String')
>>> x = s("foo")
```

At this point `s.toString` (the `JavaMultipleMethod`) has `j_self` set to `x`, but the `JavaMethod` in its `.instance_methods` does not. Which seems inconsistent to me.
Since calling `x.toString()` will invoke `__get__` (which sets `j_self` on the JMM) and `__call__` (which sets `j_self` on the JM) this is not *currently* triggerable.

PR #569, opened 2020-10-22
set_resolve_info: replace j_self w/ resolve_static

### LINKED ISSUE #557: set_resolve_info passes None instead of j_self

obfusk:
Fixes #557.

Replaces #568.

In `set_resolve_info`, `j_self` was really only used to distinguish between whether or not to resolve static or instance methods.
Using a boolean makes a lot more sense IMHO (and `.j_self` is still set by `__get__`).

### DIFF
--- a/jnius/jnius_export_class.pxi
+++ b/jnius/jnius_export_class.pxi
@@ -224,16 +224,17 @@ class MetaJavaClass(MetaJavaBase):
         # search all the static JavaMethod within our class, and resolve them
         cdef JavaMethod jm
         cdef JavaMultipleMethod jmm
+        cdef jboolean resolve_static = True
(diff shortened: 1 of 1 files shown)


cmacdonald:
Could there be a comment detailing how static differs from resolve_static?

obfusk:
@cmacdonald Better?

tshirtman (approved):
LGTM

PR #595, opened 2021-06-18
Use Python standard library `which` instead of OS `which`

dogatekin:
When `pyjnius` is installed in an environment that doesn't have the `which` command, the two functions `get_jdk_home` and `get_jre_home` might fail if they try to invoke the OS `which` command.

This PR fixes this issue by instead using the `which` implementation in the Python standard library, namely `shutil.which` for Python 3 and `distutils.spawn.find_executable` for Python 2. If the searched executable wasn't found in the original code the setup was failing with an error, so similarly I raised an explanatory exception in those cases.

If the conditional import seems icky, we could:
- Drop the `find_executable` import if you decide to drop Python 2 support, or
- Drop the `shutil.which` import since `find_executable` also works on Python 3, I only kept it because [`shutil.which`](https://github.com/python/cpython/blob/main/Lib/shutil.py#L1381) seems to have some minor improvements over [`find_executable`](https://github.com/python/cpython/blob/main/Lib/distutils/spawn.py#L95) e.g. checking `PATHEXT` for Windows.

### DIFF
--- a/jnius/env.py
+++ b/jnius/env.py
@@ -16,6 +16,11 @@ log = getLogger(__name__)
 
 PY2 = sys.version_info.major < 3
 
+if PY2:
+    from distutils.spawn import find_executable as which
+else:
+    from shutil import which
+
 machine = machine()  # not expected to change at runtime
 
 # This dictionary converts values from platform.machine()
@@ -102,11 +107,12 @@ def get_jre_home(platform):
         jre_home = join(JAVA_HOME, 'jre')
 
     if platform != 'win32' and not jre_home:
-        jre_home = realpath(
-            check_output(
-                split('which java')
-            ).decode('utf-8').strip()
-        ).replace('bin/java', '')
+        try:
+            jre_home = realpath(
+                which('java')
+            ).replace('bin/java', '')
+        except TypeError:
+            raise Exception('Unable to find java')
 
     if platform == 'win32':
         if isinstance(jre_home, bytes):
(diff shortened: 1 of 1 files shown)


cmacdonald:
Hi @dogatekin there is a similar PR under review at https://github.com/kivy/pyjnius/pull/541 - does this address your problem? Or could the patch be rebased for that?

cmacdonald:
Python 2 support is also being removed, AIUI

dogatekin:
Hi @cmacdonald, thanks for checking my changes and for pointing me to the other PR!

The PR at #541 might indeed solve the problem I've been running into since it looks like the `which` calls would be avoided as long as one of `JAVA_HOME` / `JDK_HOME` / `JRE_HOME` is set. And I could definitely rebase this patch on that PR, or we could add this change to that PR as well.

However, since that PR has been open for so long and since @tshirtman doesn't currently have that much time to fully review and test it, I thought I might have a better chance fixing this problem earlier with this more specific PR that only changes this small part, while #541 has more major changes. My ideal scenario would be to merge this PR quickly, and #541 can get merged when it receives the necessary attention and review. And since that PR still retains the calls to the system `which`, we could rebase it on these changes easily if this PR indeed gets merged quicker.

cmacdonald:
Sorry I hadnt realised you were on #541 as well!

tshirtman (approved):
Looks good to me!
I didn’t test it, but `which` seems to do exactly what we need, so i don’t see any reason to block this.
and TIL about it, so thanks :)

PR #597, opened 2021-07-14
Fix test ignored exceptions

tshirtman:
Fix broken tests + change exception for type mismatches from JavaException to TypeError

### DIFF
--- a/jnius/jnius_conversion.pxi
+++ b/jnius/jnius_conversion.pxi
@@ -34,7 +34,7 @@ cdef void release_args(JNIEnv *j_env, tuple definition_args, pass_by_reference, 
                     pass
             j_env[0].DeleteLocalRef(j_env, j_args[index].l)
 
-cdef void populate_args(JNIEnv *j_env, tuple definition_args, jvalue *j_args, args):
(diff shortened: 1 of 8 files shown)


tshirtman:
@cmacdonald maybe? Any qualms about changing the type of the exception? I feel TypeError is more appropriate than JavaException here, but i guess we could have a subclass of both in case users still want to distinguish.

cmacdonald:
I'm supportive of a change to TypeError for type mismatches. In general, it would be handy to separate Jnius problems from actual Java exceptions generated _inside_ Java.  I think there are also problems about multiple JNI startups, altering classpath after startup  etc. that should be e.g. JniusError rather than JavaException

In particular, when inside-Java exceptions occur, I would like to propagate more of the exception into the Python exception object (for easier debugging).

cmacdonald (approved):

tshirtman:
Thanks, indeed, more exception types would be useful to discriminate between the various possible issues.

I’m sometime tempted by RuntimeError for the kind of issues you describe, but it’s too broad a basket, so a JniusError could make sense.

PR #603, opened 2021-09-02
Use platform.machine() as default get_cpu() return value

Cheaterman:
It's much more likely to be correct, eg on aarch64 in my case. :-)

### DIFF
--- a/jnius/env.py
+++ b/jnius/env.py
@@ -349,5 +349,5 @@ def get_cpu():
             "WARNING: Not able to assign machine()"
             " = %s to a cpu value!" % machine
         )
-        print("         Using cpu = 'i386' instead!")
-        return 'i386'
+        print(f"         Using cpu = '{machine}' instead!")
+        return machine
--- a/jnius/env.py
+++ b/jnius/env.py
@@ -31,6 +31,7 @@ MACHINE2CPU = {
     "x86_64": "amd64",
     "AMD64": "amd64",
     "armv7l": "arm",
+    "aarch64": "aarch64",
     "sun4u": "sparcv9",
     "sun4v": "sparcv9"
 }


cmacdonald:
Fair, but are there some entries missing in MACHINE2CPU?

Cheaterman:
> Fair, but are there some entries missing in MACHINE2CPU?
> i.e. my recollection is that MACHINE2CPU maps the return of platform.machine() to the equivalent CPU type that the JDK uses in its filename.

Yup - this is all because aarch64 is missing, but since it maps to the same name in JDK path I figured this fix would be better, and could perhaps work by default with other platforms too?

cmacdonald:
For reference:

platform.machine() reports the uname system call: possible values https://stackoverflow.com/a/45125525

Discussion of Java values:
http://mail.openjdk.java.net/pipermail/jigsaw-dev/2017-April/012107.html

I support using platform.machine() as a default, but I think MACHINE2CPU should be updated for aarch64 to eliminate the warning for that platform.

Cheaterman:
> For reference:
> 
> platform.machine() reports the uname system call: possible values https://stackoverflow.com/a/45125525
> 
> Discussion of Java values:
> http://mail.openjdk.java.net/pipermail/jigsaw-dev/2017-April/012107.html
> 
> I support using platform.machine() as a default, but I think MACHINE2CPU should be updated for aarch64 to eliminate the warning for that platform.

I somehow completely ignored the warning but you're obviously right! I'll also take a look at your links to see if I can get some extra MACHINE2CPU entries in the PR :-)

Cheaterman:
I checked your links and didn't see anything obvious to add - at least on the Java side of things. I suppose this will cover most platforms for now :-)

cmacdonald (approved):
LGTM

tshirtman (approved):

Issue #618, opened 2022-03-01
“python_requires” should be set with “>=3.6”, as pyjnius 1.4.1 is not compatible with all Python versions.


PyVCEchecker:
Currently, the keyword argument **python_requires** of **setup()** is not set, and thus it is assumed that this distribution is compatible with all Python versions. 
However, I found it is not compatible with **Python<3.6**. My local Python version is 3.5, and I encounter the following error when executing _“pip install pyjnius”_ 

```
Collecting pyjnius
  Downloading pyjnius-1.4.1.tar.gz (47 kB)
     |████████████████████████████████| 47 kB 232 kB/s 
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  ERROR: Command errored out with exit status 1:
   command: /usr/local/bin/python /usr/local/lib/python3.5/site-packages/pip/_vendor/pep517/_in_process.py get_requires_for_build_wheel /tmp/tmp61sq3ndw
       cwd: /tmp/pip-install-4yqajmr2/pyjnius
  Complete output (21 lines):
  Traceback (most recent call last):
    File "/usr/local/lib/python3.5/site-packages/pip/_vendor/pep517/_in_process.py", line 280, in <module>
      main()
    File "/usr/local/lib/python3.5/site-packages/pip/_vendor/pep517/_in_process.py", line 263, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
    File "/usr/local/lib/python3.5/site-packages/pip/_vendor/pep517/_in_process.py", line 114, in get_requires_for_build_wheel
      return hook(config_settings)
    File "/tmp/pip-build-env-wavx3f29/overlay/lib/python3.5/site-packages/setuptools/build_meta.py", line 150, in get_requires_for_build_wheel
      config_settings, requirements=['wheel'])
    File "/tmp/pip-build-env-wavx3f29/overlay/lib/python3.5/site-packages/setuptools/build_meta.py", line 130, in _get_build_requires
      self.run_setup()
    File "/tmp/pip-build-env-wavx3f29/overlay/lib/python3.5/site-packages/setuptools/build_meta.py", line 254, in run_setup
      self).run_setup(setup_script=setup_script)
    File "/tmp/pip-build-env-wavx3f29/overlay/lib/python3.5/site-packages/setuptools/build_meta.py", line 145, in run_setup
      exec(compile(code, __file__, 'exec'), locals())
    File "setup.py", line 24, in <module>
      from env import (
[...]

PR #619, opened 2022-03-12
Remove Python 2 support, six dependency

rdb:
Thanks for this great project!
 
Python 2 is ancient history by now, and as pointed out in #618, pyjnius does not even work with Python versions under 3.6 at this point.

This PR removes the compatibility hacks for Python 2 and drops the now unnecessary dependency on the `six` package.

### DIFF
--- a/jnius/__init__.py
+++ b/jnius/__init__.py
@@ -44,15 +44,13 @@ else:
     from .jnius import *  # noqa
     from .reflect import *  # noqa
 
-from six import with_metaclass
-
 # XXX monkey patch methods that cannot be in cython.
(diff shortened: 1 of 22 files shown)


misl6:
We're in the process of releasing version `1.4.2`, which will be a bugfix one.
As this one will break Python 2 compatibility, I've added it to milestone `1.5.0`.

@rdb We now have some conflicts, can you take care of it?

rdb:
Sorry for the delay, I've addressed the merge conflicts and rebased.  Tests pass locally on Linux.

misl6 (approved):
LGTM! Thank you!

PR #622, opened 2022-05-23
Add suffix to support IBM jre on Windows

Tomahaawk:
`jvm.dll` is located at `/bin/default` if you are using jdk from IBM Semeru, so jnius was not finding it. Adding the suffix to where jnius searches the dll fixes the issue.

For reference, you can find the IBM Semeru jdk in this link: https://developer.ibm.com/languages/java/semeru-runtimes/downloads/

### DIFF
--- a/jnius/__init__.py
+++ b/jnius/__init__.py
@@ -21,8 +21,10 @@ if sys.platform == 'win32' and sys.version_info >= (3, 8):
         for suffix in (
             ('bin', 'client'),
             ('bin', 'server'),
+            ('bin', 'default'),
             ('jre', 'bin', 'client'),
             ('jre', 'bin', 'server'),
+            ('jre', 'bin', 'default'),
         ):
             path = os.path.join(jdk_home, *suffix)
             if not os.path.isdir(path):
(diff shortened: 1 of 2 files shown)


cmacdonald:
does the fix only apply to the Windows platform?

Tomahaawk:
Yes, you are correct @cmacdonald, the fix only applies to Windows platform. I've been using Windows as my main platform, I have not tried it in a different one.

cmacdonald:
Does this JRE support platforms other than Windows though?

Tomahaawk:
Linux and macOS are also supported

misl6 (approved):
LGTM. Thank you!

I will add "on Windows" suffix before merge on your PR title, to keep it clear that this fix is targeting only the Windows platform.

Tomahaawk:
Thanks a lot @misl6! Do you guys happen to have an expected date for a new release?

misl6:
Hi @Tomahaawk , A new release was expected yesterday but due to some issues with the CD pipeline I had to postpone it until I got it fixed. I expect that the new release (1.4.2) will be shipped before the end of this week 😃

PR #627, opened 2022-06-12
PyPy: Fixes a segfault + add tests

misl6:
- Add test rules for PyPy
- Fixed a test that may fail on PyPy (not on CI as `python` is available as an alias)
- Disabled tests for x86 on PyPy
-  Removes usage (was intended as a temp fix) of `PyUnicode_DecodeUTF16`, which was segfaulting on PyPy (Fixes on Cython have been introduced here: https://github.com/cython/cython/commit/5c9b32c3169f26174ff5a31f8a9fbe9b920933f3)

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -3,6 +3,7 @@ name: Continuous Integration
 jobs:
   Tests:
     name: base
+    continue-on-error: true
     strategy:
       matrix:
         python:
@@ -11,6 +12,9 @@ jobs:
            - '3.8'
            - '3.9'
            - '3.10'
+           - 'pypy-3.7'
+           - 'pypy-3.8'
+           - 'pypy-3.9'
         java:
           - '8'
           # - '9' # commented out just for faster CI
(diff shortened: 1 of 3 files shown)


misl6:
@tshirtman , maybe you want to give it a look before the merge?

tshirtman (approved):
LGTM :)
Edit: we don't currently set a minimum version of cython in the pyproject.toml, maybe it's a good time to do so, asking for the version with that fix or newer?

misl6:
> LGTM :)
> Edit: we don't currently set a minimum version of cython in the pyproject.toml, maybe it's a good time to do so, asking for the version with that fix or newer?

Good point! I guess is better to ask for `Cython>=0.29.30` (as is the latest available, it contains the needed fixes, and seems to work great).

BTW, I see some improvements (regarding `Cython` version) that could be made to the CI , `pyproject.toml` and `setup.py`. But, may go outside of the scope of this PR, so looks cleaner to submit a subsequent PR (seems easy and should land in a couple of hours). (I will refer to this PR, so we can keep track of it)

PR #633, opened 2022-07-26
Add BSD Unix build support (FreeBSD, OpenBSD, NetBSD, ..).

cederom:
* Builds fine on FreeBSD.
* OpenBSD returns build errors on `build_ext()` call.

Signed-off-by: Tomasz 'CeDeROM' CEDRO <tomek@cedro.info>

### DIFF
--- a/jnius/env.py
+++ b/jnius/env.py
@@ -86,6 +86,8 @@ def get_java_setup(platform=DEFAULT_PLATFORM):
         return WindowsJavaLocation(platform, JAVA_HOME)
     if platform == "darwin": #only this?
         return MacOsXJavaLocation(platform, JAVA_HOME)    
+    if 'bsd' in platform:
+        return BSDJavaLocation(platform, JAVA_HOME)
     if platform in ('linux', 'linux2', 'sunos5'): #only this?
         return UnixJavaLocation(platform, JAVA_HOME)
     log.warning("warning: unknown platform %s assuming linux or sunOS" % platform)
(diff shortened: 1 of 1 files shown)


tshirtman (approved):
Looks good to me, wondering if the "in bsd" test might be casting too wide a net, but can’t think of any platform that would be impacted, so that’s probably just fine.

cederom:
Thank you @tshirtman :-)

This is the initial patch to add build support for all BSD. I am working on FreeBSD and this enables build and install :-) It may need some tuning for other BSD flavors (like OpenBSD, NetBSD, NomadBSD, MidnightBSD, etc) but the road is open now :-)

`sys.platform` returns `freebsd13` on my FreeBSD desktop and `openbsd71` on OpenBSD thus `bsd` only check and removal of the release number :-)

I have created OpenBSD VM just to verify. It fails on extension built telling me that `.xpi` file is unknown. If you have some spare time we can track this issue now and fix.. or leave it for some OpenBSD users when interest shows up? Up to You :-)

tshirtman:
No reason to block it, it can be improved later when interrest shows up indeed :).

Thanks!

cederom:
Thank you @tshirtman :-)

Allright, OpenBSD build problem was caused by missing `cython` package, after `pip install cython` inside virtualenv build goes fine :-)

Maybe we could add `requirements.txt` file with all (venv) python modules required for development? :-)

tshirtman:
Cython is declared as a build dependency in pyproject.toml, so if the build is done in a standard way (we use cibuildwheels in the github action), then it should be installed in the build virtualenv.

How did you build it?

PR #639, opened 2022-10-15
Build aarch64 wheels

janaknat:
Successful builds: https://github.com/janaknat/pyjnius/actions/runs/3305376283

### DIFF
--- a/.github/workflows/create.yml
+++ b/.github/workflows/create.yml
@@ -14,7 +14,7 @@ jobs:
       CIBW_REPAIR_WHEEL_COMMAND_MACOS: ''
     strategy:
       matrix:
-        os: ['ubuntu-latest', 'windows-latest', 'macos-latest']
+        os: ['windows-latest', 'macos-latest']
         architecture: ['x64']
         include:
           - os: windows-latest
@@ -25,9 +25,18 @@ jobs:
             cibw_skip: '*-win32'
           - os: ubuntu-latest
             architecture: 'x64'
-            cibw_skip: '*-manylinux_i686'
+            cibw_archs: 'x86_64'
+          - os: ubuntu-latest
+            architecture: 'aarch64'
+            cibw_archs: aarch64
     runs-on: ${{ matrix.os }}
     steps:
+      - name: Set up QEMU for aarch64 wheels
+        if: ${{ matrix.cibw_archs == 'aarch64' }}
+        uses: docker/setup-qemu-action@v1
+        with:
+          platforms: arm64
+
       - name: Checkout pyjnius
         uses: actions/checkout@v3
 
@@ -47,9 +56,21 @@ jobs:
           distribution: 'temurin'
           architecture: ${{ matrix.architecture }}
 
-      - name: Build wheels
+      - name: Build Windows and MacOS wheels
+        if: ${{ matrix.os != 'ubuntu-latest' }}
         env:
           CIBW_SKIP: '*musllinux* ${{ matrix.cibw_skip }}'
+        run: python -m cibuildwheel --output-dir dist
+
+      - name: Build and test Linux wheels
+        if: ${{ matrix.os == 'ubuntu-latest' }}
+        env:
+          CIBW_SKIP: '*musllinux* ${{ matrix.cibw_skip }}'
+          CIBW_ARCHS: '${{ matrix.cibw_archs }}'
+          CIBW_TEST_REQUIRES: 'pytest'
+          CIBW_TEST_EXTRAS: 'dev,ci'
+          CIBW_BEFORE_TEST: 'yum install ant -y && cd {project} && ant all'
+          CIBW_TEST_COMMAND: 'cd {project}/tests/ && CLASSPATH=../build/test-classes:../build/classes python -m pytest -v'
         run: python -m cibuildwheel --output-dir dist
 
       - name: upload wheels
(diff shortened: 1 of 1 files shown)


misl6:
Love that feature!

@janaknat , I see some duplicated code from `.github/workflows/create.yml`, can you merge your feature into the `create.yml` file in order to avoid duplication?

janaknat:
@misl6 Done. Moved the code to `.github/workflows/create.yml`.

.github/workflows/create.yml
```diff
@@ -28,6 +28,12 @@ jobs:
             cibw_skip: '*-manylinux_i686'
     runs-on: ${{ matrix.os }}
     steps:
+      - name: Set up QEMU for aarch64 wheels
+        if: runner.os == 'Linux'
+        uses: docker/setup-qemu-action@v1
+        with:
+          platforms: arm64
```
misl6, line 35: QEMU only needs to be set up for `aarch64` builds, not also `x86_64` ones.
We're doing something similar here: https://github.com/kivy/kivy/blob/master/.github/workflows/manylinux_wheels.yml

(Same applies for the `Build and test aarch64 wheels`) step.

misl6 (changes requested):

janaknat:
@misl6 Added checks.

.github/workflows/create.yml
```diff
@@ -48,8 +58,19 @@ jobs:
...
+        run: python -m cibuildwheel --output-dir dist
+
+      - name: Build and test aarch64 wheels
+        if: ${{ matrix.cibw_archs == 'aarch64' }}
         env:
           CIBW_SKIP: '*musllinux* ${{ matrix.cibw_skip }}'
+          CIBW_ARCHS: 'aarch64'
+          CIBW_TEST_REQUIRES: 'pytest'
+          CIBW_BEFORE_TEST: 'yum install ant -y && cd {project} && ant all'
+          CIBW_TEST_COMMAND: 'cd {project}/tests/ && CLASSPATH=../build/test-classes:../build/classes python -m pytest -v'
```
misl6, line 73: Optimization: we can expand the tests also to the `x86_64` version, so there's no need to keep a separate step for `aarch64` wheels.

janaknat:
@misl6 Both x86 and aarch64 Linux wheels are built and tested in a single step.

janaknat:
@misl6 Ping!

janaknat:
@misl6 Ping!

.github/workflows/create.yml
```diff
@@ -25,9 +25,21 @@ jobs:
             cibw_skip: '*-win32'
           - os: ubuntu-latest
             architecture: 'x64'
-            cibw_skip: '*-manylinux_i686'
+            cibw_archs: 'x86_64'
+          - os: ubuntu-latest
+            architecture: 'x86'
+            cibw_archs: 'i686'
```
misl6, line 31: We want to skip the build for `*-manylinux_i686` wheels, as we removed support in a previous version.
Can you please change the config to only build `x86_64` and `aarch64` wheels?

.github/workflows/create.yml
```diff
@@ -47,11 +59,23 @@ jobs:
...
+      - name: Build and test Linux wheels
+        if: ${{ matrix.os == 'ubuntu-latest' }}
+        env:
+          CIBW_SKIP: '*musllinux* ${{ matrix.cibw_skip }}'
+          CIBW_ARCHS: '${{ matrix.cibw_archs }}'
+          CIBW_TEST_REQUIRES: 'pytest'
+          CIBW_TEST_EXTRAS: 'dev,ci'
+          CIBW_BEFORE_TEST: 'yum install ant -y && cd {project} && ant all'
+          CIBW_TEST_COMMAND: 'cd {project}/tests/ && CLASSPATH=../build/test-classes:../build/classes python -m pytest -v'
+        run: python -m cibuildwheel --output-dir dist
```
misl6, line 77: Can we keep the tests separate from the build phase?

misl6 (changes requested):
Hi @janaknat ,

I'm sorry for the late reply, I've left a couple of comments that need to be addressed before the merge.

Thank you for your support!

.github/workflows/create.yml
janaknat: Previously, it was built in this step and tested later on in a separate step. We want to use that approach? This current approach is to build the wheels and if successful, the CIBW_TEST_COMMAND is run to test the wheels.

.github/workflows/create.yml
janaknat: Sure. Will remove i686 (x86).

.github/workflows/create.yml
misl6: Yep, but unfortunately (as shown here: https://github.com/pypa/cibuildwheel/blob/main/README.md#how-it-works), tests are run into the same container / runner session.
Keeping it separate allows us to be sure that the build scripts/installations are not influencing the tests.
(A new and clean `virtualenv` is created and used for tests, but it's not enough sometime)

janaknat: We can test the wheels in a separate step for x86_64. For aarch64, that won't be possible because github natively does not support arm64 architecture.

misl6: For `aarch64` tests, `qemu` will help 😀

When setting up the test for `aarch64`, we can add:

```
- name: Set up QEMU for aarch64 tests
  if: ${{ matrix.test_arch == 'aarch64' }} # here we need to add `test_arch` in the matrix.
  uses: docker/setup-qemu-action@v1
  with:
      platforms: arm64
```

janaknat: @misl6 True, but that only sets up the QEMU for aarch64 binary. Any commands run further will be on the host architecture. To run the tests, we will have to use a docker image (manylinux2014_aarch64), setup python, pytest, ant. Then install the generated wheels, run 'ant all' and then the unit tests. cibuildwheels sets all of that up for us when we specify what we need using CIBW_TEST_REQUIRES. Unfortunately, cibuildwheel does not have a 'skip build and test only' option. We'll probably end up re-implementing the testing logic of cibuildwheel ourselves, unless there is another option.

janaknat: @misl6 Any suggestions?

misl6: Hi @janaknat !
We're doing something similar here (that is actually used to build dependencies, as we can cache them, but the same approach could be used to process tests): https://github.com/kivy/kivy/blob/a94505ef854acae0046ab8d822a4c96c011bd269/.github/workflows/manylinux_wheels.yml#L73-L77

janaknat:
@misl6 Thank you for helping with this!

Issue #663, opened 2023-07-24
pyjnius build error on s390x


ravigumm:
In pyjnius 1.5.0, with cython 0.29.35, I am getting the following error when trying to install pyjnius from source code on s390x `unknown file type '.pxi'`
```
pyjnius]# pip install .
Processing /root/ravi/pyjnius
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Building wheels for collected packages: pyjnius
  Building wheel for pyjnius (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
  × Building wheel for pyjnius (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [27 lines of output]
      warning: [options] bootstrap class path not set in conjunction with -source 1.7
      1 warning
      running bdist_wheel
      running build
      running build_py
      creating build
      creating build/lib.linux-s390x-cpython-310
      copying jnius_config.py -> build/lib.linux-s390x-cpython-310
      copying setup_sdist.py -> build/lib.linux-s390x-cpython-310
      creating build/lib.linux-s390x-cpython-310/jnius
      copying jnius/__init__.py -> build/lib.linux-s390x-cpython-310/jnius
      copying jnius/env.py -> build/lib.linux-s390x-cpython-310/jnius
      copying jnius/reflect.py -> build/lib.linux-s390x-cpython-310/jnius
      copying jnius/signatures.py -> build/lib.linux-s390x-cpython-310/jnius
      creating build/lib.linux-s390x-cpython-310/jnius/src
      creating build/lib.linux-s390x-cpython-310/jnius/src/org
      creating build/lib.linux-s390x-cpython-310/jnius/src/org/jnius
      copying jnius/src/org/jnius/NativeInvocationHandler.java -> build/lib.linux-s390x-cpython-310/jnius/src/org/jnius
      copying jnius/src/org/jnius/NativeInvocationHandler.class -> build/lib.linux-s390x-cpython-310/jnius/src/org/jnius
      running build_ext
      warning: jnius/config.pxi:1:0: The 'DEF' statement is deprecated and will be removed in a future Cython version. Consider using global variables, constants, and in-place literals instead. See https://github.com/cython/cython/issues/4310
      warning: jnius/jnius.pyx:102:0: The 'IF' statement is deprecated and will be removed in a future Cython version. Consider using runtime conditions or C macros instead. See https://github.com/cython/cython/issues/4310
[...]

PR #665, opened 2023-08-03
Force cython 2.x

cmacdonald:
As noted in #663, Jnius does not build currently. This is because of the release last month of Cython 3.0.

There are two issues:
(1) Cython 3.0 wont compile include files (.pxi). I have excluded these
(2) Jnius compiled using Cython 3.0 fails with `AttributeError: 'Class' object has no attribute '_JavaClass__cls_storage'`. I cant easily see why this is occurring.

I think we should do a new 1.5.1 release as soon as possible using this PR, as anyone installing Jnius in an environment without prebuilt binaries will currently not succeed. Pegging to Cython <3.0 will address the main problem for now. We can create another issue for Cython 3 support.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -70,6 +70,7 @@ jobs:
     - name: Install pyjnius with [dev, ci] extras
       run: |
         pip install --timeout=120 .[dev,ci]
+        pip list
   
     - name: (Windows) Test pyjnius via pytest
       if: matrix.os == 'windows-latest'
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -70,7 +70,6 @@ jobs:
     - name: Install pyjnius with [dev, ci] extras
       run: |
         pip install --timeout=120 .[dev,ci]
-        pip list
   
     - name: (Windows) Test pyjnius via pytest
       if: matrix.os == 'windows-latest'
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,5 +2,5 @@
 requires = [
     "setuptools>=58.0.0",
     "wheel",
-    "Cython>=0.29.30"
+    "Cython>=0.29.30,<3"
 ]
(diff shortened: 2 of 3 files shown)


misl6:
These `.pxi` and `.pxd` files should not be completely ignored, otherwise, if a change is done on a `.pxd` or `.pxi` file, but the `.pyx` file is left untouched, the build will not be performed.

But, we all agree that `.pxi` and `.pxd` files should not be considered sources.

Instead, an approach, similar to the one we currently use on `pyobjus` should be used:

https://github.com/kivy/pyobjus/blob/9d66904623df5e0772dad807a6a14c340f22fc9e/setup.py#L122-L128

(Tests are passing on Cython 3.x, so IMHO there's no need to force Cython 0.x)

Yeah, have a lot of deprecation warnings, but that will be a slow migration process for the whole Python community.

cmacdonald:
> (Tests are passing on Cython 3.x, so IMHO there's no need to force Cython 0.x)

Are you sure that tests and installation is passing/working on Cython 3? Its definitely not my recent experience, even on the Github Actions for this PR.

I'm happy to revert setup.py if we are forcing old Cython. 

>Instead, an approach, similar to the one we currently use on pyobjus should be used:

I think your point is to use a depends= kwarg to declare the .pxi files, right?

cmacdonald:
Any thoughts @misl6 ?

misl6:
> I think your point is to use a depends= kwarg to declare the .pxi files, right?

Yes.

> Are you sure that tests and installation is passing/working on Cython 3? Its definitely not my recent experience, even on the Github Actions for this PR.

I've tested it locally by using the `depends=` approach for `.pxi` and `.pxd` files, and everything seems to pass (so the only issue with Cython3 seems that `.pxi` and `.pxd` files should not be placed into the `sources` list)

cmacdonald:
have added `depends=` kwarg as requested. 

I, respectfully, disagree with the assessment on Cython 3. I have separated that into a separate PR, based on this PR, for which the github actions are failing.

cmacdonald:
Closing this PR in favour of #669.

PR #672, opened 2023-09-19
Support Java 20, remove Java 7 support

cmacdonald:
Java 20 (released in March 2023) has removed support for compiling class files for JDK 7. This means Jnius doesn’t currently compile (nor run?) under Java 20.

Java SE 7 had End of Extended Support in July 2022. So we need no longer support Java 7 as a target.

Please consider this PR for 1.6.0 release cycle that
- supports and tests Java 20 in GHA
- Removes Java 7 as the target

NB: I had to change the Java distribution in GHA to one that had Java 20 on most platforms.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -45,6 +45,8 @@ jobs:
         - os: windows-latest
           architecture: 'x86'
           python: 'pypy-3.9'
+        - os: windows-latest
+          java: '20'
 
     runs-on: ${{ matrix.os }}
     steps:
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -18,7 +18,7 @@ jobs:
         java:
           - '8'
           - '11'
-          - '17'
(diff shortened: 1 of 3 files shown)


misl6:
Hi @cmacdonald !

`temurin` seems to have Java `20` for all the platforms, except Windows x86. (Windows x86_64 is supported)

https://github.com/cmacdonald/pyjnius/actions/runs/6230818141/job/16911405627 seems to fail for another reason, that you later addressed.

Skipping the test on Windows x86 only seems much better than skipping it on both (we can even remove support for Windows x86 in future)

What are your thoughts?

cmacdonald:
Makes sense to me. Are you able to push to the PR to make the change.

misl6:
> Makes sense to me. Are you able to push to the PR to make the change.

Thank you for speeding up the process! Seems to work!

LGTM. Thank you!

misl6 (approved):

PR #677, opened 2023-10-01
Now Github Actions provides python3 via setup-python also for Apple Silicon Macs

misl6:
Finally, no more apple-silicon specific workarounds! 🎉 (at least on selected Python3 versions, but that's enough for `kivy-ios`)

Similar PRs: https://github.com/kivy/kivy-ios/pull/829, https://github.com/kivy/buildozer/pull/1687


cmacdonald:
Questions:
 - Why is `os: apple-silicon-m1` in test_wheels but not build?
 - Shouldnt push.yml also have `os: apple-silicon-m1` ?

PR #678, opened 2023-10-03
test for python 3.12

cmacdonald:


cmacdonald:
python 3.12 not available yet under actions/setup-python. this can wait for later.

misl6 (approved):
`3.12` is now available on Github Actions, and the CI is ✅.

I guess we should add `3.12` support on the upcoming release.

LGTM. Thank you!

cmacdonald:
should we reduce testing of interim versions of Python - e.g. just oldest and newest?

misl6:
> should we reduce testing of interim versions of Python - e.g. just oldest and newest?

Mmm, that could reduce the CI running time, but do we really want to take this risk?

What we certainly do, is remove tests for `python: 3.7` since reached EOL on June 2023.

cmacdonald:
I would keep Python 3.7 builds for longer...

One possible way to look at it - do we expect interactions between OS and python version. Or could we test most python versions on Linux, and have a more focussed CI strategy for other platforms?

cmacdonald:
PS: should setup.py specify the minimum python version:
https://stackoverflow.com/a/45362425

cmacdonald:
Should the build wheels also include Python 3.12?

misl6:
> I would keep Python 3.7 _builds_ for longer... And if you are building, you should test, right?
> 
> One possible way to look at it - do we expect interactions between OS and python version. Or could we test most python versions on Linux, and have a more focussed CI strategy for other platforms?

We can try, but I do not feel it's something super-great. IIRC, on `pypy-x` we had a specific `pypy` version that was failing for apparently no reason on macOS, and that version was neither the oldest or the newest one.

... Then the issue has been fixed "automagically", so I can't really tell you why, but I feel more comfortable to test along all the versions.

But, I'm open for a discussion (on a separate thread)

misl6:
> Should the build wheels also include Python 3.12?

Absolutely. We need to update `cibuildwheels` for that.

misl6:
> PS: should setup.py specify the minimum python version: https://stackoverflow.com/a/45362425

That's a good idea.

PR #684, opened 2023-10-08
Add (now mandatory) `.readthedocs.yaml` file, add docs extras and update sphinx conf

misl6:
readthedocs.org [now requires](https://blog.readthedocs.com/migrate-configuration-v2/) a `v2` configuration file, in order to build the docs.

Unfortunately, for this reason, readthedocs was unable to build docs for latest `1.6.0` pyjnius release. (so a `1.6.1` will be released just after this PR is merged)

This PR does the following:
- Adds a `.readthedocs.yaml` file, with the proper configuration.
- Adds a `requirements.txt` file into `docs` folder, so `readthedocs.org` can use that to install deps.
- Updates the theme to [furo](https://github.com/pradyunsg/furo), a nice and clean theme, also used by `pip`.
- Takes the `pyjnius` version from `pyjnius`, by reading it on `__init__.py` file (we never updated that previously, so manual is not an option 😀, but we can't install `pyjnius`)
- Now the copyright year is updated automatically  (again: we never updated that previously, so manual is not an option 😀)

For PR reviewer,  the newly rendered docs are available here: https://pyjnius.readthedocs.io/en/feat-add-readthedocs-file/


misl6:
Update: Moved back in WIP, as while doing the same changes for `pyobjus`, I needed to follow another approach (as we can't install `pyobjus` on readthedocs), and would be nice to use the same.

misl6:
Update: Now `readthedocs.org` can process the docs build without installing `pyjnius` (and its dependencies)

cmacdonald:
lgtm

AndreMiras (approved):
Thanks for investigating into this

Issue #695, opened 2023-11-08
installation documentation could be refreshed


cmacdonald:
https://pyjnius.readthedocs.io/en/latest/installation.html talks a lot about installing Cython etc, which isnt really necessary. You can pip install the binaries. That content should move to a building page, and make the installation much easier.

cmacdonald:
Seems reasonable.

Was wondering about separating installation to another page tbh, but if it's clear that it's not normally needed, then it's already an improvement.

misl6:
Fixed via #700

PR #700, opened 2023-11-20
Refresh installation and build docs

cmacdonald:
This addresses the changes discussed in #695.

### DIFF
--- a/docs/source/building.rst
+++ b/docs/source/building.rst
@@ -1,29 +1,30 @@
 .. _building:
 
-Building PYJNIus
+Building PyJNIus
 ================
 
 Building PyJNIus is necessary for development purposes, or if there is no 
 pre-built binary for your particular platform. 
 
-Like installation of PyJNIus, building PyJNIus requires a  `Java
-<https://www.oracle.com/java/technologies/downloads/>`_ Development Kit 
+Like installation of PyJNIus, building PyJNIus requires a  `Java Development Kit
+<https://www.oracle.com/java/technologies/downloads/>`_  (JDK)
 to be installed.
 
-Apart from the JDK, the building requirements for each platform are as follows:
+Apart from the JDK, the build requirements for each platform are as follows:
 
- - Linux: the GNU Compiler Collection (GCC).
- - Windows Microsoft Visual C++ Build Tools (command-line tools subset of Visual
-   Studio) can be obtained from https://visualstudio.microsoft.com/downloads/. 
+ - Linux: the `GNU Compiler Collection <https://gcc.gnu.org/>`_ (GCC), e.g. using 
+   `apt-get install build-essentials` on Debian-based distributions.
+ - Windows: `Microsoft Visual C++ Build Tools <https://visualstudio.microsoft.com/downloads/>`_ 
+   (the command-line tools subset of Visual Studio). 
    For more information or options, see Python's `Windows Compilers wiki
    <https://wiki.python.org/moin/WindowsCompilers>`_.
  - macOS: `Xcode command-line tools <https://mac.install.guide/commandlinetools/index.html>`_.
 
-In all cases, PyJNIus can be *built* and installed using::
+In all cases, after checking out PyJNIus from GitHub, it can be *built* and installed using::
 
      pip install .
 
-This installs Cython (as specified in the 
+This installs `Cython <https://cython.org/>`_ (as specified in the 
 `pyproject.toml <https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/>`_) 
 in a Python build environment. On all platforms, if PyJNIus cannot find your JDK, you can set 
 the `JAVA_HOME` shell environment variable (this is often needed on Windows).
--- a/docs/source/building.rst
+++ b/docs/source/building.rst
@@ -4,8 +4,8 @@ Building PYJNIus
 ================
 
 Building PyJNIus depends on `Cython <http://cython.org/>`_ and the `Java
-<http://www.oracle.com/javase>`_ Development Kit (includes the Java Runtime
-Environment).
+<https://www.oracle.com/java/technologies/downloads/>`_ Development Kit 
+(includes the Java Runtime Environment).
 
 
 Building on GNU/Linux distributions
@@ -33,7 +33,7 @@ Python and pip must be installed and present in the ``PATH`` environment variabl
 
 1. Download and install the JDK containing the JRE:
 
-   http://www.oracle.com/technetwork/java/javase/downloads/index.html
+   https://www.oracle.com/java/technologies/downloads/
 
 2. Edit your system and environment variables (use the appropriate Java bitness
    and version in the paths):
@@ -86,7 +86,7 @@ Python and pip must be installed and present in the ``PATH`` environment variabl
 
 1. Download and install the JDK containing the JRE:
 
-   http://www.oracle.com/technetwork/java/javase/downloads/index.html
+   https://www.oracle.com/java/technologies/downloads/
 
 2. Edit your system and environment variables (use the appropriate Java bitness
    and version in the paths):
--- a/docs/source/building.rst
+++ b/docs/source/building.rst
@@ -3,20 +3,32 @@
 Building PYJNIus
 ================
 
-Building PyJNIus depends on `Cython <http://cython.org/>`_ and the `Java
+Building PyJNIus is necessary for development purposes, or if there is no 
+pre-built binary for your particular platform. 
+
+Like installation of PyJNIus, building PyJNIus requires a  `Java
 <https://www.oracle.com/java/technologies/downloads/>`_ Development Kit 
-(includes the Java Runtime Environment).
+to be installed.
 
+Apart from the JDK, the building requirements for each platform are as follows:
 
-Building on GNU/Linux distributions
------------------------------------
+ - Linux: the GNU Compiler Collection (GCC).
+ - Windows Microsoft Visual C++ Build Tools (command-line tools subset of Visual
+   Studio) can be obtained from https://visualstudio.microsoft.com/downloads/. 
+   For more information or options, see Python's `Windows Compilers wiki
+   <https://wiki.python.org/moin/WindowsCompilers>`_.
+ - macOS: `Xcode command-line tools <https://mac.install.guide/commandlinetools/index.html>`_.
 
-You need the GNU Compiler Collection (GCC), the JDK installed (openjdk
-will do), and Cython. Then, just type::
+In all cases, PyJNIus can be *built* and installed using::
 
-    pip install .
+     pip install .
 
-If you want to compile the extension within the directory for any development,
+This installs Cython (as specified in the 
+`pyproject.toml <https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/>`_) 
+in a Python build environment. On all platforms, if PyJNIus cannot find your JDK, you can set 
+the `JAVA_HOME` shell environment variable (this is often needed on Windows).
+
+If you want to compile the PyJNIus extension within the directory for any development,
 just type::
 
     make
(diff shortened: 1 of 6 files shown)


Julian-O (commented):
* File name contains typo: build**L**ing.rst

docs/source/buildling.rst
```diff
@@ -0,0 +1,160 @@
...
+
+Building PYJNIus
+================
+
+Building PyJNIus depends on `Cython <http://cython.org/>`_ and the `Java
+<http://www.oracle.com/javase>`_ Development Kit (includes the Java Runtime
+Environment).
+
+
+Installation on GNU/Linux distributions
```
Julian-O, line 11: [Here and throughout] The subsections are titled "Installation" but should be titled "Building"

cmacdonald:
Thanks for the review @Julian-O. Feedback addressed.

docs/source/installation.rst
```diff
@@ -3,28 +3,22 @@
 Installation
 ============
 
-PyJNIus depends on `Cython <http://cython.org/>`_ and the `Java
-<http://www.oracle.com/javase>`_ Development Kit (includes the Java Runtime
-Environment).
+Pyjnius has pre-compiled binaries on PyPi for recent Python versions on Linux,
```
misl6, line 6: Minor:
```suggestion
PyJNIus has pre-compiled binaries on PyPi for recent Python versions on Linux,
```

docs/source/installation.rst
```diff
@@ -3,28 +3,22 @@
...
 
+On each platform::
 
-Installation on GNU/Linux distributions
----------------------------------------
+   pip install pyjnius
 
-You need the GNU Compiler Collection (GCC), the JDK and JRE installed (openjdk
-will do), and Cython. Then, just type::
+should successfully install the platform.
```
misl6, line 13: Did you meant `package` ?

docs/source/installation.rst
```diff
@@ -3,28 +3,22 @@
...
-
-You can run the tests suite to make sure everything is running right::
-
-    make tests
+If there is no pre-compiled binary available, pip install will aim to compile
+a binary on your operating system. For more information see the :ref:`building` 
+documentation.
 
+You will need the Java JDK installed (`OpenJDK <https://openjdk.org/>`__` is fine).
+Pyjnius searches for Java in the usual places on each operating system. If PyJNIus
```
misl6, line 20: Minor:
```suggestion
PyJNIus searches for Java in the usual places on each operating system. If PyJNIus 
```

docs/source/installation.rst
```diff
@@ -3,28 +3,22 @@
...
-You can run the tests suite to make sure everything is running right::
-
-    make tests
+If there is no pre-compiled binary available, pip install will aim to compile
+a binary on your operating system. For more information see the :ref:`building` 
+documentation.
 
+You will need the Java JDK installed (`OpenJDK <https://openjdk.org/>`__` is fine).
+Pyjnius searches for Java in the usual places on each operating system. If PyJNIus 
+cannot find Java, set the JAVA_HOME environment variable.
```
misl6, line 21:
```suggestion
cannot find Java, set the `JAVA_HOME` environment variable.
```

docs/source/quickstart.rst
```diff
@@ -4,8 +4,9 @@ Quickstart
 ==========
 
 Eager to get started? This page will give you a good introduction to Pyjnius. It assumes
-you have already Pyjnius installed. If you do not, head over the
-:ref:`installation` section.
+you have already Pyjnius installed. If you do not, ensure you have a Java JDK installed,
```
misl6, line 7:
```suggestion
you have already PyJNIus installed. If you do not, ensure you have a Java JDK installed, 
```

docs/source/building.rst
```diff
@@ -0,0 +1,123 @@
...
+Environment).
+
+
+Building on GNU/Linux distributions
+-----------------------------------
+
+You need the GNU Compiler Collection (GCC), the JDK installed (openjdk
+will do), and Cython. Then, just type::
+
+    sudo python setup.py install
```
misl6, line 17: Ouch! We really want to suggest using `sudo` ? 😬

And, maybe is better to suggest using `pip install .` ?

docs/source/building.rst
```diff
@@ -0,0 +1,123 @@
...
+
+   http://www.oracle.com/technetwork/java/javase/downloads/index.html
+
+2. Edit your system and environment variables (use the appropriate Java bitness
+   and version in the paths):
+
+    Add to your `Environment Variables
+    <https://en.wikipedia.org/wiki/Environment_variable>`_:
+
+    * ``JAVA_HOME``: C:\\Program Files\\Java\\jdk1.7.0_79\\bin
```
misl6, line 45: Maybe is better to keep the path more generic, without suggesting a specific (outdated) one?

docs/source/building.rst
```diff
@@ -0,0 +1,123 @@
...
+      http://aka.ms/vcpython27
+
+   b) MinGWPy for Python 2.7:
+
+      https://anaconda.org/carlkl/mingwpy
+
+   c) Microsoft Visual C++ Build Tools (command-line tools subset of Visual
+      Studio) for Python 3.5 and 3.6:
+
+      https://visualstudio.microsoft.com/downloads/
```
misl6, line 68: Ouch, that sounds a little outdated. Better to remove the `2.7`, `3.5` and `3.6` and keep the Python3 version only.

misl6 (changes requested):
Hi @cmacdonald !

I've left some comments. Some of them are not really tied to your PR, but makes sense to update what looks outdated, while we're moving things.

Let me know what you think about it.

docs/source/building.rst
```diff
@@ -0,0 +1,112 @@
...
+Building PyJNIus depends on `Cython <http://cython.org/>`_ and the `Java
+<http://www.oracle.com/javase>`_ Development Kit (includes the Java Runtime
+Environment).
+
+
+Building on GNU/Linux distributions
+-----------------------------------
+
+You need the GNU Compiler Collection (GCC), the JDK installed (openjdk
+will do), and Cython. Then, just type::
```
cmacdonald, line 15: My recollection is that doing pip creates a venv, so Cython need not be installed these days?

docs/source/building.rst
```diff
@@ -0,0 +1,112 @@
...
+
+
+Building on Windows
+-------------------
+
+Python and pip must be installed and present in the ``PATH`` environment variable.
+
+1. Download and install the JDK containing the JRE:
+
+   http://www.oracle.com/technetwork/java/javase/downloads/index.html
```
cmacdonald: Why point to Oracle here, when in installation.rst we say that OpenJDK is fine?

docs/source/building.rst
```diff
@@ -0,0 +1,112 @@
...
+
+3. Download and install the C compiler:
+
+   Microsoft Visual C++ Build Tools (command-line tools subset of Visual
+   Studio) can be obtained from https://visualstudio.microsoft.com/downloads/
+
+   For more information or options, see Python's `Windows Compilers wiki
+   <https://wiki.python.org/moin/WindowsCompilers>`_.
+
+4. Update `pip <https://pip.pypa.io/en/stable/installing>`_ and setuptools::
```
cmacdonald, line 62: Why we need this? Surely setuptools is installed if needed?

docs/source/building.rst
```diff
@@ -0,0 +1,112 @@
...
+
+      python -m pip install --upgrade pip setuptools
+
+5. Install Cython::
```
cmacdonald, line 66: As above, Cython installed using pip.

docs/source/building.rst
```diff
@@ -0,0 +1,112 @@
...
+
+You need the GNU Compiler Collection (GCC), the JDK installed (openjdk
+will do), and Cython. Then, just type::
+
+    pip install .
+
+If you want to compile the extension within the directory for any development,
+just type::
+
+    make
```
cmacdonald, line 22: Perhaps this is what requires Cython. pip install install it in the venv, right?

docs/source/building.rst
```diff
@@ -0,0 +1,112 @@
...
+
+Building for macOS
+------------------
+
+Python and pip must be installed and present in the ``PATH`` environment variable.
+
+
+1. Download and install the JDK containing the JRE:
+
+   http://www.oracle.com/technetwork/java/javase/downloads/index.html
```
cmacdonald: as above, should we point to OpenJDK? 
Or for mac, suggest using brew?

docs/source/building.rst
```diff
@@ -0,0 +1,112 @@
...
+
+2. Edit your system and environment variables (use the appropriate Java bitness
```
cmacdonald, line 91: I think this step is simply unnecessary now. Pyjnius knows to call /usr/libexec/java_home to find the Java installation

docs/source/building.rst
```diff
@@ -0,0 +1,112 @@
...
+
+    * ``export JAVA_HOME=/usr/libexec/java_home``
+
+3. Install Xcode command-line tools.
+
+4. Update `pip <https://pip.pypa.io/en/stable/installing>`_ and setuptools::
+
+      python -m pip install --upgrade pip setuptools
+
+5. Install Cython::
```
cmacdonald, line 105: As above, do we need to mention this?

docs/source/building.rst
```diff
@@ -0,0 +1,112 @@
+.. _building:
+
+Building PYJNIus
```
cmacdonald, line 3: Perhaps this document could be much simpler:

We list the requirements for BUILDING for each platform, i.e.  JDK and
  - GCC for Linux
  - Visual Studio for Windows
  - Xcode tools for macOS

Then tell people that pip install will do the rest.

cmacdonald (commented):
I read over the building.rst in more details, adding my own thoughts

docs/source/building.rst
misl6: Yeah, unlike `kivy/kivy` where we have a lot of dependencies and things to build (like SDL prerequisites), `kivy/pyjnius` has a limited set of dependencies.

And now, building `pyjnius` is definitely simpler than before.

So, I agree.

docs/source/building.rst
misl6: Yes, you're right. `pip install .` or `pip install -e .` will install build deps (like `Cython`) in a separate build-only environment, so it's not needed anymore.

docs/source/building.rst
misl6: This requires Cython, but we can change the make script (or delete it), since now things are a way more easier to do via `pip`.

docs/source/building.rst
misl6: Yes, same of `Cython`

docs/source/building.rst
misl6: Yes.

docs/source/building.rst
misl6: Not anymore. (If we completely remove any reference to `python setup.py install`)

docs/source/building.rst
cmacdonald: addressed by mentioning Cython here.

cmacdonald:
Thanks @misl6. I vastly simplified the building page. Screenshot here for ease of reference:

image

Perhaps the Visual Studio link could be inline like the rest?

Julian-O:
Quick style review of screenshot:

* Title has inconsistent capitalisation.
* Java Development Kit - make the link apply to the whole phrase, not just Java.
* Java Development Kit - add "(JDK)" to introduce the abbreviation that is used below.
* "the building requirements" -> "the build requirements"
* Gnu Compiler Collection - add a link to https://gcc.gnu.org/ - consider adding the name of the apt-get package.
* Yes, inline the Microsoft downloads link
* `pip install .` - I never know whether to specify that you need to download the code, and change into the directory with setup.py before running this - I mean, the sort of people who want to do this know that, right? right??
* "This installs Cython" - add a link to what Cython is
* "shell environment variable" - remove the word "shell" which is a Unix-specific term.

cmacdonald:
Revised version:
image

> "shell environment variable" - remove the word "shell" which is a Unix-specific term.

I currently left this, as the previous sentence talks about Python build environments.

Julian-O:
> > "shell environment variable" - remove the word "shell" which is a Unix-specific term.
> 
> I currently left this, as the previous sentence talks about Python build environments.

I don't care enough to suggest a further change, but both Windows and Linux build environments have "environment variables", but only Linux (and Unix) use the term "shell" so when you talk about "shell environment variables" being needed on Windows, it doesn't match up.

misl6 (approved):
LGTM. Thank you!

Agree with @Julian-O regarding env variables, but I also do not want to lock this PR progress. Feel free to address it separately.

Issue #699, opened 2023-11-18
Usage without JDK


ennioVisco:
Hello,

I cannot find docs/other issues on this topic, so I considered using a new issue.

In recent versions of Java, tools likeJava modules and `jpackage` made it extremely easily to build tailor-made JREs that select only the parts of the JDK that are used by the target Java library.

I was thinking that it would be really cool if there was a way to use that JRE, to avoiding completely the extra step for the user of installing and switching to the appropriate version of the JDK.

Why is this not possibile? Are there known challenges in that?

cmacdonald:
I had some thoughts about this. 

1. Pyjnius allows access to all of Java, so limiting to particular modules probably precludes. 

2. IMO, any packaging of minimal Java for a particular Python application that use jnius seems out of scope for the Pyjnius project itself.

3. The JDK is definitely needed for _building_ Pyjnius. I haven't thought about the choice of JRE/JDK for _using_ Pyjnius - but I have a feeling for the JDK is needed.

ennioVisco:
> I had some thoughts about this.

Thanks for the nice comments, let me share my thoughts and ideas about it:

### My comments on your nice thoughts

> 1. Pyjnius allows access to all of Java, so limiting to particular modules probably precludes.

I definitely agree with this, I don't think Pyjnius should be limited in this sense.



> 2. IMO, any packaging of minimal Java for a particular Python application that use jnius seems out of scope for the Pyjnius project itself.

I also agree with this, I don't believe necessarily that Pyjnius should be concerned by packaging, in my view, the ideal setting would just be to allow the user of Pyjnius to select a JRE that is different from the one at JAVA_HOME!



> 3. The JDK is definitely needed for _building_ Pyjnius. I haven't thought about the choice of JRE/JDK for _using_ Pyjnius - but I have a feeling for the JDK is needed.

I'm sorry, I confess my lacks in understanding how Pyjnius works. While I think it makes perfect sense that the full JDK is needed _when building_, at the same time I would expect Pyjnius's __runtime requirements__ from the jdk to be some subset of the JDK, and, specifically, the union between the set of JDK modules needed by the running application + some set of extra JDK modules required by Pyjnius itself (say, e.g. the Reflection API). 
I would consider reasonable to include these dependencies at build time in Java, and therefore an ad-hoc runtime should work without any issues.

### The use case I have in mind
[...]

PR #705, opened 2023-12-21
Support installing and testing using only a JRE 

cmacdonald:
#699 asked if a JDK was really required, or if a JRE is sufficient.

This creates a new Github Action that tests if the JREs (rather than JDKs) are sufficient to run and test pyjnius. I started from create.yml, but had to add a new stage that uses ant to obtain the `build/test-classes` folder.

The answer appears to be positive.

Note that building jnius still requires a JDK. I've updated the documentation.

### DIFF
--- a/.github/workflows/create.yml
+++ b/.github/workflows/create.yml
@@ -152,6 +152,7 @@ jobs:
         with:
           java-version: '17'
           distribution: 'temurin'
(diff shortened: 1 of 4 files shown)


cmacdonald:
Not sure the relevance of JRE for kivy-ubuntu-arm64. @misl6?

PR #706, opened 2023-12-22
require minimum python version in pyproject.toml

cmacdonald:
noted in #618
i chose to change pyproject.toml - is this better than changing setup.py?

(actually SETUP_KWARGS in https://github.com/kivy/pyjnius/blob/master/setup_sdist.py#L26)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,6 @@
+[project]
+requires-python = ">= 3.7"
+
 [build-system]
 requires = [
     "setuptools>=58.0.0",


pyproject.toml
```diff
@@ -1,3 +1,7 @@
+[project]
+name = 'pyjnius'
+requires-python = ">= 3.7"
```
cclauss, line 3: Pythonista 3.9 will end-of-life this week.  Perhaps this should be 3.10 so users are not building solutions on a slower and less secure base.

PR #707, opened 2024-01-19
Use `pathlib` to get path to `site-packages` instead of `importlib.resources`

occasional-contributor:
The existing method fails with `ValueError` when using `pyjnius` in a [`slash`](https://github.com/getslash/slash) test.


cmacdonald:
I'm the originator of this code, which follows the recommendations in the documentation linked in the code. 

Can you say more about the raised ValueError? Why is importlib.resources not the best way?

cmacdonald:
please.

PR #708, opened 2024-01-28
[build wheel] Use the new Ubuntu ARM64 runner

misl6:
`kivy-ubuntu-22.04-arm64`, is a newer scalable Ubuntu ARM64 self-hosted runner managed by us, and hosted on Azure (thanks Microsoft).


misl6:
Closed as now GitHub provides hosted arm64 ubuntu runners.

PR #711, opened 2024-02-07
Update quickstart.rst

Suletta-Majo:
minimum sample's forgot bracket fix
below document is sourced from here? so i fixed it from referenced github's code
https://pyjnius.readthedocs.io/en/latest/quickstart.html


Julian-O:
PEP8 would suggest that there should be no space between the function name ('print') and the first parenthesis.

Julian-O (approved):

Issue #736, opened 2024-11-19
BUG: failing `test_hierharchy_arraylist` with jdk>=21


h-vetinari:
When [building](https://github.com/conda-forge/pyjnius-feedstock/pull/43) pyjnius 1.6.1 against newer openjdk (21, 22) in conda-forge, things compile and run fine, with the exception of one test failure:
```
=================================== FAILURES ===================================
____________________ ReflectTest.test_hierharchy_arraylist _____________________

self = <test_reflect.ReflectTest testMethod=test_hierharchy_arraylist>

    def test_hierharchy_arraylist(self):
        d = identify_hierarchy_dict(find_javaclass("java.util.ArrayList"), 0, True)
        self.assertContains(d, "java.util.ArrayList")# concrete
        self.assertContains(d, "java.util.AbstractCollection")# superclass
        self.assertContains(d, "java.util.Collection")# interface
        self.assertContains(d, "java.lang.Iterable")# interface
        self.assertContains(d, "java.lang.Object")# root
        maxLevel = max(d.values())
>       self.assertEqual(d["java.lang.Object"], maxLevel)
E       AssertionError: 3 != 4

tests/test_reflect.py:38: AssertionError
=========================== short test summary info ============================
FAILED tests/test_reflect.py::ReflectTest::test_hierharchy_arraylist - AssertionError: 3 != 4
======================== 1 failed, 164 passed in 0.94s =========================
```

It seems all the expected hierarchy levels are still there, but either the hierarchy itself changed for real (meaning the test expectation is wrong), or the calculation in `identify_hierarchy_dict` needs to fix something to take into account the new reality.

cmacdonald:
i can confirm the tests failing and, I think, the original author of this test. I'll prepare a PR; getting the maintainers to merge it will be another matter!

cmacdonald:
Java 18:
```python
{
'java.lang.Object': 3, 
'java.lang.Iterable': 3, 
'java.util.Collection': 2,
'java.util.AbstractCollection': 2, 
'java.util.List': 1, 
'java.util.AbstractList': 1, 
'java.util.RandomAccess': 1, 
'java.lang.Cloneable': 1, 
'java.io.Serializable': 1, 
'java.util.ArrayList': 0
}
```
Java 21:
```python
{
'java.lang.Object': 3, 
'java.lang.Iterable': 4, 
'java.util.Collection': 3, 
'java.util.AbstractCollection': 2, 
[...]

PR #741, opened 2025-01-24
fix for BUG: failing test_hierharchy_arraylist with jdk>=21 #736

cmacdonald:
As reported by @h-vetinari the test test_hierharchy_arraylist made assumptions about the hierarchy of objects that are no longer relevant. This softens the test so it works on jdks 21-23.

Before merging, we need to agree on the JDK versions we should be testing. Windows termurin seems to only have upto 18?


tshirtman:
As far as i can tell, the change should work on the previous versions as well, as it accepts the result we used to enforce. I'm not sure it should be conditional on changing JDK versions tested in CI. Dropping old versions is always a tradeoff, and it can be acceptable, even if some software still didn't upgrade their support.

looking at
https://endoflife.date/oracle-jdk
https://endoflife.date/redhat-build-of-openjdk
and https://endoflife.date/microsoft-build-of-openjdk

doesn't seem to give any consistent indication to what should be supported…

h-vetinari:
Thanks for working on a fix! This will fix #736

h-vetinari:
> doesn't seem to give any consistent indication to what should be supported…

Plus there are alternative implementations of the java runtime; a good overview of the timelines can be found at [wikipedia](https://en.wikipedia.org/wiki/Java_version_history#Release_table). Of course, pyjnius could choose not to support all versions anymore, many projects in the ecosystem have dropped Java 8 already, for example.

tshirtman (approved):

cmacdonald:
Thanks @misl6 !

PR #745, opened 2025-02-26
Create devcontainer.json

rusty-svg:
Java implemention


cmacdonald:
What is this for? Why is it needed?

rusty-svg:
I try developomg an own cluster compute System with specified Prozess an Operation models and Architecture aß my first programming project.

PR #752, opened 2025-05-10
Fix building from source failure because of Cython 3.1.0

younes-makhchan:
The recent release of Cython 3.1.0 broke the building and installation of pyjnius from source

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,5 +2,5 @@
 requires = [
     "setuptools>=58.0.0",
     "wheel",
-    "Cython"
+    "Cython==3.0.12"
 ]


younes-makhchan:
@tshirtman  @AndreMiras  @kollivier  @artagnon  can some please look into this

cmacdonald:
jnius builds with Cython < 3 too...
Can you post another issue with the exact error message you are seeing, so we can have an issue to fix this, in addition to/instead of pegging to a particular Cython version.

mmadhuhasa:
Hi,
Below images shows the error occurs during the docker image build,
please try to fix this issue ASAP or accept this pull request which is working.

image
image

pyproject.toml
```diff
@@ -2,5 +2,5 @@
 requires = [
     "setuptools>=58.0.0",
     "wheel",
-    "Cython"
+    "Cython==3.0.12"
```
AndreMiras, line 5: Why don't we `Cython<3.1.0` rather than hard pinning to a specific version?

cmacdonald:
i've provided a PR in #753 that addresses the underlying problem

pyproject.toml
younes-makhchan: just at that time that worked for me, but actually your idea is good too !

PR #756, opened 2025-06-16
Remove python long to support Cython >= 3.1.x

cmacdonald:
In https://github.com/kivy/pyjnius/pull/753#issuecomment-2969400311 it was suggested by @da-woods that the fix for #753 was sub-optimal:

> `ctypes.c_long` probably isn't the right fix. `isinstance(x, long)` was testing for the Python 2 long type (which no longer exists in Python 3 - Python 3 just has one integer type that covers both large and small integers).

This PR:
- removes `ctypes.c_long`
- when scoring method signatures, it better tests argument value ranges for compatibility for both short and integer types
- adds some more test cases
- adds a `debug` kwarg for method and constructor invocation that displays the scores of the different methods that are being scored in `calculate_scores()`. This may be useful when use of the `signature` kwarg is necessary.

Example output:
```
>>> clz = autoclass("org.jnius.SignatureTest$IntOrLong")
>>> obj = clz(0, debug=True) # could be int or long
[(20, '(I)V', 'V', ('I',), (0,)), (20, '(J)V', 'V', ('J',), (0,))]
Selected (J)V for invocation
```
(ie. constructors (I)V and (J)V both got score of 20, jnius selected (J)V.

### DIFF
--- a/jnius/jnius_conversion.pxi
+++ b/jnius/jnius_conversion.pxi
@@ -38,7 +38,6 @@ cdef void populate_args(JNIEnv *j_env, tuple definition_args, jvalue *j_args, ar
     cdef JavaClass jc
     cdef PythonJavaClass pc
     cdef int index
-    from ctypes import c_long as long
 
     for index, argtype in enumerate(definition_args):
         py_arg = args[index]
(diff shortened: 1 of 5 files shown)


cmacdonald:
> All failures are Pypy 3.8, which Cython 3.1 may not support?

We already looked at this in #752: see https://github.com/kivy/pyjnius/pull/753#issuecomment-2889148812

T-Dynamos:
This PR is required for https://github.com/kivy/python-for-android/pull/3180.

cmacdonald:
rebased!

misl6 (approved):
LGTM. Thank you!

misl6:
@T-Dynamos + @cmacdonald , I guess we're only missing Python 3.13 support (and maybe Cython < 3) removal to process a release?

T-Dynamos:
@misl6 Yes, I believe this is ready for release. I've also tested this PR on Android.

PR #765, opened 2025-07-28
Migrate to `download-artifact@v4` and `upload-artifact@v4`

misl6:
This pull request updates the GitHub Actions workflow in `.github/workflows/create.yml` to use the latest versions of actions. The key changes include upgrading `upload-artifact` and `download-artifact` actions to version `v4` and introducing dynamic artifact naming as uploading multiple artifacts with the same name is not supported anymore.

See: https://github.com/actions/download-artifact/blob/main/docs/MIGRATION.md

### DIFF
--- a/.github/workflows/create.yml
+++ b/.github/workflows/create.yml
@@ -68,9 +68,9 @@ jobs:
           python -m cibuildwheel --output-dir dist
 
       - name: upload wheels
-        uses: actions/upload-artifact@v2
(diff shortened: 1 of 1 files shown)


cmacdonald:
mirrorlist.centos.org...  https://serverfault.com/a/1161847

cmacdonald:
I wonder if we need to update [cibuildwheel](https://github.com/pypa/cibuildwheel)

PR #771, opened 2025-08-15
Fix all Cython conditional keywords

kmedya-dev:
- Replace IF with if, ELIF with elif, and ELSE with else
- Ensures proper inclusion of platform-specific .pxi files
- Prevents build errors due to invalid Cython syntax

### DIFF
--- a/jnius/jnius.pyx
+++ b/jnius/jnius.pyx
@@ -99,11 +99,11 @@ include "jnius_compat.pxi"
 include "jni.pxi"
 include "config.pxi"
 
-if JNIUS_PLATFORM == "android":
+IF JNIUS_PLATFORM == "android":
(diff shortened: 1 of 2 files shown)


cmacdonald:
which Cython version is causing this?  We're trying to keep track of Cython upgrades...

kmedya-dev:
Cython version=3.1.3(letest version)

cmacdonald:
I think the error is elsewhere - thats just a warning.

Have you tried installing the latest github master - we made various Cython related fixes, see history at https://github.com/kivy/pyjnius/commits/master/

cmacdonald:
can you paste the whole output? you have to hunt around for the actual build error in the build logs

PR #780, opened 2025-09-04
Make env.py publically accessible, without starting the JVM

cmacdonald:
I have a usecase where I wish to know the JDK version before starting Jnius.
For instance, JDK24+ will need certain options to suppress warnings about native things.

I thought it would be handy to get the JDK location from jnius, but *before* jnius starts the JVM.

This PR moves env.py into jnius_config, so its accessible before jnius starts.

If thought necessary for backwards compatibility, I can make a jnius/env.py that has deprecates imports from jnius_config/env.py?


misl6 (approved):
LGTM. Thank you!

